PR #95, opened 2026-04-09
Improve search: multi-term AND + relevance ranking (FTS spike)

### LINKED ISSUE #84: Explore DuckDB FTS extension for full-text search in Explorer
## Problem

The Interactive Explorer currently uses `ILIKE '%term%'` for text search ([line 398](https://github.com/isamplesorg/isamplesorg.github.io/blob/main/tutorials/isamples_explorer.qmd#L398)). This is a brute-force substring scan — no ranking, no stemming, no stopword handling.

## Proposal

Use DuckDB's [full-text search extension](https://duckdb.org/docs/current/core_extensions/full_text_search), which is [supported in DuckDB-WASM](https://duckdb.org/docs/current/clients/wasm/extensions.html). This would give us BM25-ranked results, Porter stemming, and stopword filtering.

## Approach: Pre-built index

Building the FTS index at runtime in the browser would be too slow for 6.7M records. Instead:

1. **Build offline**: Run `PRAGMA create_fts_index('samples', 'pid', 'label', 'description', 'place_name')` in native DuckDB
2. **Export as `.duckdb` file**: The FTS index is stored as plain tables + macros in a named schema
3. **Host on `data.isamples.org`** alongside the parquet files
4. **`ATTACH` in browser**: `ATTACH 'https://data.isamples.org/isamples_fts_index.duckdb'` — BM25 scoring comes along for free

### Query change

```sql
-- Before (substring scan):
WHERE label ILIKE '%pottery%' OR description ILIKE '%pottery%'

-- After (ranked FTS):
SELECT *, score
FROM (
  SELECT *, fts_main_samples.match_bm25(pid, 'pottery') AS score
  FROM samples
)
WHERE score IS NOT NULL
ORDER BY score DESC
```

## Open questions

- How large will the `.duckdb` index file be for 6.7M records with 3-4 text columns?
- Does `ATTACH` over HTTP work reliably in DuckDB-WASM for files this size?
- Should we index all text columns or just `label` + `description`?
- Alternative: [DuckDB text analytics functions](https://duckdb.org/2025/06/13/text-analytics) (stemming, stopwords) without a full FTS index — lighter weight but still better than raw `ILIKE`

## Related

- #82 — Explorer query architecture documentation
- Raised by @ekansa in Slack discussion

rdhyee:
## Summary

Closes #84 — FTS spike complete with immediate search improvements and documented future path.

**Shipped now (zero new dependencies):**
- **Multi-term search**: "pottery Cyprus" requires BOTH words to match (was OR on the full phrase)
- **Wildcard-safe ILIKE**: search terms containing `%`, `_`, or `\` are escaped, so they match literal characters instead of acting as wildcards
- **Relevance ranking**: results sorted by score when searching — label match = 3pts, place_name = 2pts. (Description is not scored — the column lives in the wide parquet, while the Explorer's search runs against the lite parquet, which has no description.)
- When not searching, results remain random for exploration variety
- Search composes with active facet filters (Material / Sampled Feature / Specimen Type) and the source legend, so results stay inside the user's current narrowing

**FTS spike findings:**
- Built offline DuckDB FTS index with `tools/build_fts_index.py` (preserved as a non-production spike artifact, clearly marked in its module docstring)
- Full index (label + description + place_name): **358 MB** — too large for auto-download
- Lite index (label + place_name only): **211 MB** — still substantial
- BM25 scoring works well (Porter stemming, English stopwords)
- `ATTACH` over HTTP in DuckDB-WASM is supported but downloading 200–358 MB is impractical

**Recommended next steps (not in this PR):**
1. Explore pre-tokenized search parquet (inverted index as parquet, much smaller)
2. Consider on-demand FTS loading behind an "Enhanced Search" toggle
3. Evaluate DuckDB text analytics functions (stemming without full index)

## Test plan

- [ ] Search "pottery" → results ranked by relevance (label matches first)
- [ ] Search "pottery Cyprus" → only samples matching BOTH words
- [ ] Search "basalt" → geological samples with label matches at top
- [ ] Search "100%" or other wildcard chars → matches the literal characters, not all rows
- [ ] Clear search → results return to random sampling
- [ ] With Material/Source/Specimen filters active → search results stay inside those filters
- [ ] `tools/build_fts_index.py` runs successfully with local parquet (spike artifact, not part of the deploy)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ models/generated/extensions/*.md
 
 # Large data files
 *.parquet
+*.duckdb
 
 # Node / Playwright
 node_modules/
--- a/tools/build_fts_index.py
+++ b/tools/build_fts_index.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+"""
+Build a DuckDB full-text search index for the iSamples Explorer.
+
+Creates a .duckdb file containing the FTS index (BM25-scored) that can
+be ATTACHed in DuckDB-WASM for ranked text search over 6.7M samples.
+
+Usage:
+    python tools/build_fts_index.py
+
+Output:
+    tools/isamples_fts_index.duckdb  (upload to data.isamples.org)
+
+Requirements:
+    pip install duckdb
+"""
+
+import duckdb
+import os
+import sys
+from pathlib import Path
+
+PARQUET_URL = "https://data.isamples.org/isamples_202601_wide.parquet"
+OUTPUT_DB = Path(__file__).parent / "isamples_fts_index.duckdb"
+
+# Local fallback for faster builds
+LOCAL_PARQUET = Path.home() / "Data/iSample/pqg_refining/zenodo_wide_2026-01-09.parquet"
+
+
+def build_fts_index():
+    # Use local file if available, otherwise remote
+    source = str(LOCAL_PARQUET) if LOCAL_PARQUET.exists() else PARQUET_URL
+    print(f"Source: {source}")
+
+    # Remove existing index file
+    if OUTPUT_DB.exists():
+        OUTPUT_DB.unlink()
+
+    con = duckdb.connect(str(OUTPUT_DB))
+
+    print("Creating samples table from parquet...")
+    con.execute(f"""
+        CREATE TABLE samples AS
+        SELECT
+            pid,
+            label,
+            COALESCE(description, '') AS description,
+            COALESCE(CAST(place_name AS VARCHAR), '') AS place_name
+        FROM read_parquet('{source}')
+        WHERE otype = 'MaterialSampleRecord'
+    """)
+
+    row_count = con.execute("SELECT COUNT(*) FROM samples").fetchone()[0]
+    print(f"Loaded {row_count:,} rows")
+
+    print("Installing and loading FTS extension...")
+    con.execute("INSTALL fts")
+    con.execute("LOAD fts")
+
+    print("Building FTS index (this may take a few minutes)...")
+    con.execute("""
+        PRAGMA create_fts_index(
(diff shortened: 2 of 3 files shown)


rdhyee:
Reviewed and pushed two small follow-ups (134aca2):

**1. ILIKE wildcard escaping.** Search input was passed into the `ILIKE` pattern with only single-quote escaping, so literal `%` or `_` in the query (e.g. `100%`, `co_op`) silently became wildcards. Now escape `% _ \` and add `ESCAPE '\'` in both the `whereClause` block and the relevance-score expression.

**2. FTS spike script header.** `tools/build_fts_index.py` told readers to "upload to data.isamples.org" but per the PR's own findings the 200-358 MB result is too large to ship. Reframed as `STATUS: spike artifact — NOT in production pipeline`, kept the script for future revisits, dropped the misleading upload instructions.

**Smoke test** (`/tmp/explorer_smoke_test.py` against local Quarto render):

```
Serving docs on :64856
URL: http://127.0.0.1:64856/tutorials/isamples_explorer.html
JS exceptions:    0
Console errors:   0
Failed requests:  0
RESULT: PASS
```

Exercised: initial load, multi-term search (`pottery cyprus`), wildcard-char search (`100%`). Screenshot confirms the new placeholder and that `100%` no longer matches everything.

**Other notes from review (not blocking):**
- Score expression has discrete plateaus (0/1/2/3/5/6 per term); ties break alphabetically on `label`. Fine for spike — could mention in placeholder docs later.
- `description ILIKE` over the wide parquet over HTTP range-fetch may add first-search latency; worth a `?perf=1` measurement before declaring search "done", but out of scope here.

LGTM to merge once you've eyeballed the diff.

Issue #148, opened 2026-04-28
Facets show vocabulary URIs instead of human-readable labels (cross-repo)


rdhyee:
## Problem

On the [Interactive Explorer](https://isamples.org/tutorials/isamples_explorer.html), the Material / Sampled Feature / Specimen Type facets render raw vocabulary URIs as their labels:

> `https://w3id.org/isample/vocabulary/sampledfeature/1.0/pasthumanoccupationsite (1,059,025)`
> `https://w3id.org/isample/biology/biosampledfeature/1.0/Animalia (188,361)`

The Source facet is fine (short codes — SESAR, OPENCONTEXT, GEOME, SMITHSONIAN). Everything that comes through the SKOS vocabularies is broken.

## Root Cause

The bug lives upstream of the Explorer. Sampling the wide parquet:

```sql
SELECT row_id, label, scheme_name
FROM read_parquet('https://data.isamples.org/isamples_202601_wide.parquet')
WHERE otype = 'IdentifiedConcept'
LIMIT 5;
```
returns:
```
(19401546, 'https://w3id.org/isample/vocabulary/materialsampleobjecttype/1.0/wholeorganism', NULL)
(19401547, 'https://w3id.org/isample/opencontext/materialsampleobjecttype/0.1/tile',         NULL)
(19401548, 'https://w3id.org/isample/opencontext/materialsampleobjecttype/0.1/clothing',     NULL)
```

So `IdentifiedConcept.label` already holds the URI rather than the SKOS `prefLabel`, and `scheme_name` is NULL. By the time `pqg facet-summaries` runs (`pqg/__main__.py @ aefd465`), it does:

```sql
SELECT 'material' AS facet_type, c.label AS facet_value, c.scheme_name AS scheme, COUNT(*) AS count
FROM (...) JOIN (... otype = 'IdentifiedConcept') c ON c.row_id = s.material_id
```
and faithfully copies the URI through into `facet_summaries.parquet`. The Explorer's OJS code (`tutorials/isamples_explorer.qmd:119-174`) renders `r.value` directly.

## Why a Quick Fix Isn't Right

Easy options on the table:
- **In-page string mangling** (`x.split('/').pop()`): renders `pasthumanoccupationsite`, not `Past Human Occupation Site`. Loses fidelity. Doesn't help notebooks or future consumers.
[...]

PR #149, opened 2026-04-28
Add vocab_labels.parquet builder (#148 step 1)

rdhyee:
First step of #148.

## Summary

- Adds \`scripts/build_vocab_labels.py\` — parses the same 10 SKOS TTLs that \`scripts/generate_vocab_docs.sh\` already pulls (core + Earth Sci + OpenContext + Biology), emits a single \`vocab_labels.parquet\` with columns: \`uri, uri_form, pref_label, lang, scheme, definition, alt_labels, source_ttl\`.
- 537 rows / 535 unique URIs / 10 schemes / \`en\` + \`de\`.
- **Coverage** against the 55 distinct \`IdentifiedConcept\` URIs in the current wide parquet: **51/55 (93%) by URI, >99.99% sample-weighted on every facet** (material, context, object_type).

## What it doesn't do (yet)

- Publish to \`data.isamples.org\`.
- Wire any consumer (Explorer, notebooks, pqg facet-summaries) to use it.

Both follow-ups, in their own PRs once we agree on the artifact shape.

## Two surprises handled (full write-up on the issue)

1. **\`/1.0/\` URI mismatch.** TTLs declare concepts without a version segment, but iSamples *export records* carry \`/1.0/\` in the URI. Confirmed: the version segment is in the export itself, propagated faithfully by pqg. The builder emits dual-form rows tagged \`uri_form = "vocab" | "data_v1"\` so JOINs work on either form. Per-prefix oddities (OpenContext uses \`/0.1/\`; biology data has inconsistent casing — \`Animalia/Fungi/Plantae\` but \`bacteria/protozoa\`) are handled.
2. **Cross-vocab redeclarations.** 17 material URIs declared in 2-3 TTLs each with slightly different labels (whitespace, casing). Builder dedupes \`(uri, lang)\` preferring the canonical TTL; losers' labels preserved in \`alt_labels\`.

## Residual unresolved (4 URIs, ~169 / ~6M samples)

Upstream data-quality issues — concept URIs in source data that no TTL declares:

- \`opencontext/material/0.1/organicanimalproduct\`
- \`opencontext/material/0.1/plantmaterial\`
- \`vocabulary/specimentype/1.0/othersolidobject\` (declared as \`msot:othersolidobject\` in the wrong namespace)
- \`vocabulary/specimentype/1.0/physicalspecimen\` (likewise)

Not fixable in this artifact — flagged for #148 as design debt.

## Run / deps

\`\`\`
python scripts/build_vocab_labels.py [-o out.parquet] [--also-csv]
\`\`\`

Deps: \`rdflib pandas pyarrow\`. Not run in site CI; this is a one-shot artifact builder.

## Test plan

- [x] \`python scripts/build_vocab_labels.py -o /tmp/v.parquet\` succeeds with no warnings beyond the documented dedup.
- [x] Sanity-check resolved labels for representative facet URIs (\`Animalia\`, \`Fungi\`, \`Whole organism material sample\`, \`Past human occupation site\`, \`Gaseous material\`).
- [x] Coverage check against current wide parquet: 51/55 URIs, >99.99% sample-weighted.

### DIFF
--- a/scripts/build_vocab_labels.py
+++ b/scripts/build_vocab_labels.py
@@ -1,0 +1,299 @@
+#!/usr/bin/env python3
+"""
+Build vocab_labels.parquet from the SKOS TTL vocabularies that iSamples uses.
+
+The Explorer and Python notebooks need a stable lookup from vocabulary URIs
+(e.g. https://w3id.org/isample/vocabulary/sampledfeature/1.0/pasthumanoccupationsite)
+to human-readable labels (e.g. "Past human occupation site"). This script
+parses every SKOS TTL listed in scripts/generate_vocab_docs.sh, emits one row
+per (concept URI, language) pair, and writes a single parquet file.
+
+Output columns:
+    uri          str    Concept URI (vocab-form OR data-form — see uri_form)
+    uri_form     str    "vocab"   = URI as declared in the TTL
+                        "data_v1" = synthesized URI with "/1.0/" version
+                                    segment after the scheme root (the
+                                    convention used in iSamples export
+                                    records and downstream parquet files).
+    pref_label   str    skos:prefLabel (or rdfs:label fallback)
+    lang         str    BCP47 language tag, default "en"
+    scheme       str    skos:inScheme URI (or derived)
+    definition   str?   skos:definition (best-available language)
+    alt_labels   list   skos:altLabel values plus prefLabels from any
+                        cross-vocab redeclarations of the same URI.
+    source_ttl   str    URL of the TTL the canonical row came from.
+
+The dual-form (vocab + data_v1) emission is a workaround for a known
+mismatch: the vocabulary TTLs declare concepts without a version segment,
+but iSamples export records carry URIs with a "/1.0/" segment. See
+issue #148 for the full background.
+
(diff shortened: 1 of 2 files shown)


rdhyee:
Both Codex findings addressed in 30b3c81:

**P2 #1 — partial-fetch failures producing a successful artifact**
Default behavior is now **fail-loud** (exit 3) if any TTL source fails to fetch/parse. Since this artifact is intended for publishing, silently emitting a partial parquet is the wrong default. Added `--allow-partial` for explicit override (useful during local debugging).

Verified by injecting a bogus URL into `VOCAB_TTLS`:
- without flag → exit 3, error listed on stderr
- with `--allow-partial` → exit 0, parquet written

**P2 #2 — script deps not in repo environment**
Added `scripts/requirements.txt` pinning `rdflib>=6.3`, `pandas>=2.0`, `pyarrow>=14`. Kept separate from the site-build `../requirements.txt` (the scripts here are one-shot artifact builders, not run in CI — pulling them into every Quarto render would be noise). Updated docstring to reference it.

Run:
```
pip install -r scripts/requirements.txt
python scripts/build_vocab_labels.py -o vocab_labels.parquet
```

Thanks for the review.

PR #159, opened 2026-05-01
[codex] Phase 3: add globe URL params and multi-term search

rdhyee:
## Summary

Implements issue #156 Phase 3 on top of the merged Phase 1 work.

- adds query param read/write for q, sources, material, context, object_type, maxSamples, view, and perf
- hydrates search and filters while preserving existing hash camera/sample state
- folds in multi-term AND search with relevance ordering using the globe page query style

## Verification

- quarto render tutorials/progressive_globe.qmd
- git diff --check

Draft because phases 2, 3, and 4 are parallel PRs and may need final integration ordering.

### DIFF
--- a/tutorials/progressive_globe.qmd
+++ b/tutorials/progressive_globe.qmd
@@ -1563,6 +1563,10 @@ zoomWatcher = {
         }
     });
 
+    if (searchInput && searchInput.value.trim().length >= 2) {
+        doSearch();
+    }
+
     refreshFacetCounts();
 
     // --- Deep-link: restore selection from initial hash ---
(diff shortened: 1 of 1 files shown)


rdhyee:
Review finding: `?q=` is restored into the search box, but the search is not executed on page load. A shared URL like `...?q=pottery%20Cyprus` renders the input value while leaving the results panel unchanged until the user manually clicks Search or presses Enter. For the URL params + multi-term search behavior, the visible search state should be restored automatically.

rdhyee:
Addressed in `c06dd02`: restored `?q=` searches now run automatically after the search handlers are bound, so shared query URLs populate the visible results without requiring a manual Search click.

PR #160, opened 2026-05-01
[codex] Phase 4: add globe table view

rdhyee:
## Summary

Implements issue #156 Phase 4 on top of the merged Phase 1 work.

- adds a binary Globe/Table toggle without destroying the Cesium viewer
- renders a paginated table with bounded maxSamples control
- reuses the portable Phase 1 facet predicate for table filtering

## Verification

- quarto render tutorials/progressive_globe.qmd

Draft because phases 2, 3, and 4 are parallel PRs and may need final integration ordering.

### DIFF
--- a/tutorials/progressive_globe.qmd
+++ b/tutorials/progressive_globe.qmd
@@ -105,6 +105,109 @@ format:
     }
     .search-bar button:hover { background: #0d47a1; }
     .search-results { font-size: 12px; color: #666; padding: 4px 0; }
+    .view-toolbar {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        gap: 12px;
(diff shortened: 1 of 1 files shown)


rdhyee:
Review finding: table results can go stale across view switches. `setView()` only refreshes table data when entering table mode if `rows.length === 0`, while filter-change listeners refresh only when table mode is already active. If a user loads the table, switches back to globe, changes filters, then returns to table, the old rows are reused. Off-table filter changes should mark the table dirty or force a refresh on the next table entry.

rdhyee:
Addressed in `7fd98ec`: table view now tracks a dirty state when filters change outside table mode, and refreshes on the next table entry instead of reusing stale rows.

Issue #170, opened 2026-05-08
Explorer FTS Track 3: Offline index builder + tokenizer regression set


rdhyee:
> **Updated 2026-05-08** per Codex review round 2 on #165. Concept-label join (vocab_labels.parquet) is now an explicit input + tokenization step, not an implicit consequence of #169's contract. Build-stats artifact (`build_stats.json`) added so coverage is empirical, not narrative.

Sub-issue of #165. **Depends on #169 (search_index_v1 contract doc).**

## Goal

Implement the offline pipeline that builds the v1 search index from the **sample-centric document projection** specified in #169 / `SEARCH_INDEX_V1.md`. The projection joins across the property graph and emits per-virtual-field token rows.

### 1. Builder

- File: `tools/build_search_index.py` (new; sibling to `tools/build_fts_index.py` from PR #95, which stays as the FTS spike artifact).

- **Inputs** (v1 minimum):
  - `samples_map_lite.parquet` → `sample.label`, `sample.place_name[]` per sample
  - `sample_facets_v2.parquet` → `sample.description` per sample, plus the URI lists for `material`, `context`, `object_type`
  - `vocab_labels.parquet` → SKOS prefLabel (en) for each URI; **load-bearing**: this is what makes `concept.label` work and what dereferences `<…>/Pottery` → `pottery` for the index

- **Document-projection step**:
  1. For each sample (`pid`), gather the v1-minimum text fragments tagged by virtual field:
     - `sample.label` (1 fragment per sample)
     - `sample.description` (1 fragment per sample, sparse coverage)
     - `sample.place_name` (1+ fragments from the array)
     - `concept.label` (N fragments — one per `material` URI + one per `context` URI + one per `object_type` URI; resolve each URI through `vocab_labels.parquet` and emit the prefLabel; URIs without a prefLabel fall back to URI tail and a `concept_label_missing_pref_label` build-stat counter increments)
  2. Tokenize each fragment with the Python tokenizer (§2). Track per-(pid, field) token frequencies and doc length.
  3. Emit token rows per #169 §4 schema:
     ```
     { token, pid, field, tf, doc_len }
     ```
  4. Compute global per-token DF across all (pid, field) pairs.

- **Outputs**:
  - Hash-partitioned token-row parquets per #169 §6, written to `<output_dir>/isamples_YYYYMM_search_index_v1/`.
  - Sidecar `df.parquet` with global token DF.
  - `build_stats.json` (see §6 below).

[...]

Issue #169, opened 2026-05-08
Explorer FTS Track 2: search_index_v1 contract doc


rdhyee:
> **Updated 2026-05-08** per Codex review on #165. Major changes: refactored around a sample-centric document projection (Section 1), v1 minimum now includes dereferenced concept labels, query-time policy split from build-time tokenizer (new Section 3), quality gate hardened (Section 8). Original framing preserved in git history.

Sub-issue of #165. No code dependencies; can start in parallel with #N1 (#167).

## Goal

Land a single design doc — `SEARCH_INDEX_V1.md` — that pins the v1 substrate contract before any pipeline or query code is written. Tracks 3-5 implement and measure against this doc.

### 1. Sample search document projection

The substrate is **not** "tokenize these parquet columns." It is "tokenize a *sample-centric document* whose text fragments are joined across the property graph and tagged by their entity origin."

Each sample (`pid`) has a logical document of weighted text fragments. At build time, a join across the wide parquet (or its narrow equivalents) produces this projection; the substrate then tokenizes per fragment, tagging each token row with the **virtual field name** (entity dot field), not the source parquet column.

**v1 minimum** — the projection that ships first:

| virtual field         | source                                                              | rationale                                                  |
|-----------------------|----------------------------------------------------------------------|------------------------------------------------------------|
| `sample.label`        | `MaterialSampleRecord.label` (~6.68M coverage)                       | canonical title; near-universal                            |
| `sample.description`  | `MaterialSampleRecord.description` (~1.61M ≈ 24%)                    | sparse but high-signal where present                       |
| `sample.place_name`   | `samples_map_lite.parquet.place_name[]` (~2.21M)                     | already proven valuable in current ILIKE search            |
| `concept.label`       | `material` / `context` / `object_type` URIs dereferenced via `vocab_labels.parquet` (`pref_label`, lang=en) | **load-bearing addition**: facet URIs are near-universal but raw URIs are useless to FTS; dereferenced labels make `pottery`, `ceramic`, `basalt`, `bone`, `marine` work as the user expects |

A sample whose `material` URI is `<…>/Pottery` gets a row `{token: 'pottery', pid: ..., field: 'concept.label', tf: 1, ...}`. One row per sample per facet URI per token. Coverage: most of the dataset.

**v1.5 expansion** (post-v1, additive — no schema change):

| virtual field                       | source (Solr equivalent)                                  |
|-------------------------------------|-----------------------------------------------------------|
[...]

Issue #168, opened 2026-05-08
Explorer FTS Track 1b: Honesty fix for query-spec / live mismatch


rdhyee:
Sub-issue of #165. **Direction gated by Track 1a (#N1) measurements.**

## Background

`query-spec.qmd:225` says the live Explorer searches `label + description + place_name`, but `explorer.qmd` searches only `label + place_name` (`samples_map_lite.parquet` does not carry `description`). The mismatch is documented in #165; this issue resolves it.

## Two possible directions

**Direction A — expand live to match docs.** Swap `doSearch()` to read from `sample_facets_v2.parquet`, which carries `description` (per #165 measurements: ~3.0s cold native for `pottery`).

**Direction B — narrow docs to match live.** Update `query-spec.qmd:225` and any explorer UI copy to state the live subset is `label + place_name`. Keep the current parquet source.

## Decision rule

After #N1 lands a baseline, run the same Playwright perf-smoke against a feature-branch swap to `sample_facets_v2.parquet`:

- If browser cold latency for `pottery` ≤ **5 s** AND multi-term `pottery Cyprus` ≤ **6 s** AND filter-composed search ≤ **8 s** → ship Direction A.
- Otherwise → ship Direction B.

The thresholds are intentionally generous because this is an *interim* state, not the destination. The substrate work in tracks 2-4 sets stricter targets.

## Scope

- Branch off main, swap the parquet source in `doSearch()`, run the same benchmark from #N1, append numbers to this issue.
- Open the chosen-direction PR (A or B) and link the comparison data.
- Update `query-spec.qmd:225` either way (Direction A: add date stamp confirming live matches docs; Direction B: rewrite the line to match live).

## Acceptance

- [ ] Comparison numbers (current live vs `sample_facets_v2.parquet` swap) posted to this issue
- [ ] Direction chosen mechanically per the rule above
- [ ] PR merged with the honesty fix
- [ ] **Not framed as the FTS solution.** Issue title and PR description must call this an *interim* state.

## Out of scope

- Substrate design (#N3), pipeline (#N4), browser query prototype (#N5).

## Refs

#165, #N1

rdhyee:
## Measured (local Quarto preview, 2026-05-08)

Three datasets:

- **A0 — Baseline** (current ILIKE on \`samples_map_lite\`, \`label + place_name\`)
- **A1 — Naive LEFT JOIN swap** (search facets_v2 with description, naive LEFT JOIN to lite for coords)
[...]

Issue #167, opened 2026-05-08
Explorer FTS Track 1a: Browser perf-smoke baseline


rdhyee:
Sub-issue of #165. No upstream dependencies.

## Goal

Establish a measured browser-side baseline for the current live search (`samples_map_lite.parquet`, `label + place_name`, ILIKE) before changing anything. Subsequent tracks compare against these numbers, not against speculation.

## Scope

1. **Extend `?perf=1` instrumentation** in `explorer.qmd` to log search timings.
   - Wrap `doSearch()` (`explorer.qmd:1782-1859`) with `performance.mark('search-<id>-start')` / `performance.mark('search-<id>-end')` / `performance.measure('search', ...)`.
   - Capture: query string, terms count, results count, elapsed ms, and (best-effort) bytes transferred via `performance.getEntriesByType('resource')` filtered to the parquet URL within the search window.
   - Surface in the existing perf panel.

2. **Add a Playwright perf benchmark** in `tests/`.
   - File: `tests/test_search_perf.py` (new; sibling to `tests/test_globe.py`).
   - Run the canonical query set (below) against the deployed site at `https://isamples.org/explorer.html?perf=1`.
   - Record: cold first-search latency, warm repeat-search latency, transferred bytes, results count, whether composition with active source/facet filters works.
   - Output: `tests/search_baseline_<YYYY-MM-DD>.json` artifact for future-track comparison.

## Canonical query set (locked)

| label              | query           | expected behavior                                                       |
|--------------------|-----------------|-------------------------------------------------------------------------|
| `single-common`    | `pottery`       | many results; cold + warm timings                                       |
| `single-rare`      | `basalt`        | mid-volume results                                                      |
| `multi-term`       | `pottery Cyprus`| AND semantics; tests multi-term path                                    |
| `no-hit`           | `xyzzyqqqplugh` | zero results path                                                       |
| `wildcard-literal` | `100%`          | literal `%` handling via `escapeIlikePattern`                           |
| `wildcard-literal` | `_test`         | literal `_` handling                                                    |
| `diacritic`        | `Çatalhöyük`    | NFKC + diacritic-strip behavior under current ILIKE (likely zero hits)  |
| `composed-filter`  | `pottery` + source=OPENCONTEXT only + material=`<one URI>` | source + facet composition still works |

## Acceptance

- [ ] `?perf=1` panel shows search timings per invocation
[...]

PR #173, opened 2026-05-08
explorer: search perf-smoke baseline (#167)

rdhyee:
## Summary

Track 1a of #165 — establishes a measured browser-side baseline for the current live search before any substrate work begins.

- Instruments \`doSearch()\` with \`performance.mark\` / \`performance.measure\` and emits a structured \`isamples.search\` console event on every completion (success or error). Captures \`elapsed_ms\`, \`results_count\`, \`bytes_transfer\`, \`bytes_body\`, per-URL \`seen_urls\`, and filter composition flags.
- Search measures appended to the existing \`?perf=1\` panel.
- New \`tests/test_search_perf.py\` drives the locked canonical query set, captures the events, and writes a JSON artifact to \`tests/search_baseline_<UTC_DATE>.json\`.

No behavior change to the search itself.

## Baseline numbers (committed in this PR)

Local Quarto preview, 2026-05-08. \`tests/search_baseline_2026-05-08.json\`. All 9 canonical queries completed cleanly.

| label                        | term              | results | cold (ms) | warm (ms) |
|------------------------------|-------------------|--------:|----------:|----------:|
| single-common                | \`pottery\`         |      50 |      8683 |      1940 |
| single-rare                  | \`basalt\`          |      50 |     11824 |      2400 |
| multi-term                   | \`pottery Cyprus\`  |       0 |      5071 |      2560 |
| no-hit                       | \`xyzzyqqqplugh\`   |       0 |      4870 |      2554 |
| wildcard-pct                 | \`100%\`            |       0 |      5178 |      2574 |
| wildcard-under               | \`_test\`           |       0 |     18520 |      2529 |
| diacritic                    | \`Çatalhöyük\`      |       0 |      4415 |      2592 |
| composed-source              | \`pottery\` × OPENCONTEXT |  50 |   4964 |      1368 |
| composed-source-material     | + material[0]     |       0 |      4331 |      2106 |

**Headlines** (also posted as a comment on #167):
- Current ILIKE-on-samples_map_lite **already misses every v1 contract budget** (cold 4-12s vs ≤2s target; warm 1.4-2.6s vs ≤500ms target). The static substrate prototype must substantially beat these.
- \`pottery Cyprus\` returns **0 results** because samples_map_lite has 0 Cyprus matches in label/place_name — Cyprus only appears in description. **Direct evidence for #168 honesty-fix Direction A.**
- \`Çatalhöyük\` returns 0 — current ILIKE doesn't strip diacritics. The v1 tokenizer in #170 fixes this.
- Wildcard literals (\`100%\`, \`_test\`) return 0 cleanly without errors — \`escapeIlikePattern\` is correct.
- \`bytes_transfer = 0\` and \`seen_urls = []\` everywhere — no fetches *during* the search window. DuckDB-WASM caches parquet metadata at page load and uses HTTP range reads from cache. **Implication**: "bytes per search" is a substrate-world metric, not a current-world metric. Worth pinning in #169 budget framing.

## Canonical query set (locked in #167)

\`pottery\`, \`basalt\`, \`pottery Cyprus\`, \`xyzzyqqqplugh\`, \`100%\`, \`_test\`, \`Çatalhöyük\`, \`pottery\` × OPENCONTEXT only, \`pottery\` × OPENCONTEXT × material[0].

Each query opens a **fresh browser context** for a true file-cold measurement, then re-runs on the same page for a warm measurement.

## Open design questions (recorded on #167)

1. **Log channel**: defaulted to \`console.log(JSON.stringify(...))\`. Alternative \`CustomEvent\` deferred unless non-Playwright observers appear.
2. **Bytes-transferred**: defaulted to \`performance.getEntriesByType('resource')\` summing \`transferSize\`, falling back to \`encodedBodySize\`. Cross-origin \`transferSize\` requires \`Timing-Allow-Origin\` on R2; the empty bytes in the baseline indicate that R2 may not be setting TAO, but more likely that DuckDB-WASM caches metadata at load time. Worth verifying on a deployed preview.

## Test plan

- [x] \`?perf=1\` panel shows search timings per invocation
- [x] \`tests/test_search_perf.py\` runs the canonical set and produces a JSON artifact
- [x] Baseline numbers recorded in \`tests/search_baseline_2026-05-08.json\` and as a comment on #167
- [ ] Re-run against a deployed preview once this PR merges (numbers will differ from local-quarto-preview)
- [ ] Verify whether R2 sets \`Timing-Allow-Origin: *\` (#174 deferred)

## Out of scope

- Honesty fix (Direction A vs B) — that's #168, gated by these baseline numbers.
- Substrate design — #169.

## Refs

#165, #167, #174 (deferred concerns from review)

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1877,9 +1877,12 @@ zoomWatcher = {
             try { performance.measure(`search-${searchId}`, markStart, markEnd); } catch (e) {}
             const elapsedMs = performance.now() - tStart;
 
-            // Sum bytes transferred from data.isamples.org during the search
+            // Per-URL byte data from data.isamples.org during the search
             // window. transferSize is 0 cross-origin without Timing-Allow-Origin;
-            // fall back to encodedBodySize so we still report something.
+            // encodedBodySize is reported as a fallback. Per-URL detail (rather
+            // than just summed bytes) lets analysis post-hoc-filter concurrent
+            // fetches that are not actually attributable to the search.
+            const seenUrls = [];
             let transferBytes = 0;
             let bodyBytes = 0;
             try {
@@ -1887,6 +1890,11 @@ zoomWatcher = {
                 for (const e of entries) {
                     if (!e.name.startsWith(R2_BASE)) continue;
                     if (e.startTime < tStart || e.startTime > tStart + elapsedMs) continue;
+                    seenUrls.push({
+                        name: e.name,
+                        transfer_size: e.transferSize || 0,
+                        body_size: e.encodedBodySize || 0,
+                    });
                     transferBytes += (e.transferSize || 0);
                     bodyBytes += (e.encodedBodySize || 0);
                 }
@@ -1903,6 +1911,7 @@ zoomWatcher = {
                     elapsed_ms: Math.round(elapsedMs),
                     bytes_transfer: transferBytes,
                     bytes_body: bodyBytes,
+                    seen_urls: seenUrls,
                     has_source_filter: getActiveSources().length !== SOURCE_VALUES.length,
                     has_facet_filter: hasFacetFilters(),
                     error: errorMessage,
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1612,6 +1612,24 @@ zoomWatcher = {
         }, 250);
     }
 
+    // --- Busy-flag depth counter (#173 review round 2) ---
+    //
+    // body.classList 'explorer-busy' tracks "any change-triggered async
+    // work in flight." Without depth counting, overlapping handlers race:
+    // a fast handler's `finally` removes the class while a slower
+    // handler's loadRes / facet recompute is still running, defeating the
+    // whole point of the flag. Depth-counted: class is added on the
+    // 0 → 1 transition and removed on the 1 → 0 transition.
+    let _busyDepth = 0;
+    function busyAcquire() {
+        if (_busyDepth === 0) document.body.classList.add('explorer-busy');
+        _busyDepth++;
+    }
+    function busyRelease() {
+        _busyDepth = Math.max(0, _busyDepth - 1);
+        if (_busyDepth === 0) document.body.classList.remove('explorer-busy');
+    }
+
     // --- Source filter change handler ---
     //
     // The body.classList 'explorer-busy' flag wraps every async work path
@@ -1624,7 +1642,7 @@ zoomWatcher = {
     // .recomputing-clear poll downstream is meaningful. See #173 review.
     const resUrls = { 4: h3_res4_url, 6: h3_res6_url, 8: h3_res8_url };
     document.getElementById('sourceFilter').addEventListener('change', async () => {
-        document.body.classList.add('explorer-busy');
+        busyAcquire();
         try {
             updateSourceLegendState();
             writeQueryState();
@@ -1638,7 +1656,7 @@ zoomWatcher = {
             refreshFacetCounts();
             await new Promise(r => setTimeout(r, 300));
         } finally {
-            document.body.classList.remove('explorer-busy');
+            busyRelease();
         }
     });
 
@@ -1651,7 +1669,7 @@ zoomWatcher = {
     // takes effect at neighborhood zoom. See issue #156, Phase 1.
     const facetNote = document.getElementById('facetNote');
     async function handleFacetFilterChange() {
-        document.body.classList.add('explorer-busy');
+        busyAcquire();
         try {
             const active = hasFacetFilters();
             if (facetNote) facetNote.style.display = (active && mode === 'cluster') ? 'block' : 'none';
@@ -1663,7 +1681,7 @@ zoomWatcher = {
             refreshFacetCounts();
             await new Promise(r => setTimeout(r, 300));
         } finally {
-            document.body.classList.remove('explorer-busy');
+            busyRelease();
         }
     }
     document.getElementById('materialFilterBody').addEventListener('change', handleFacetFilterChange);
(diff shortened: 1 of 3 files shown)


rdhyee:
## Self-review fixes applied

Force-pushed (rebased onto upstream main + 1 fix-up commit). PR is now 2 commits, scope is purely \`explorer.qmd\` (+72/-1) and \`tests/test_search_perf.py\` (+263).

**Dropped from PR**: the unrelated \`provenance/scancode-2026-05-04.json\` commit that had been sitting unpushed on local main; carried in by accident when this branch was cut from main. Available on local main for a separate provenance PR.

**Code fixes**:

- \`explorer.qmd\`: capture per-URL \`seen_urls\` list in the structured search log (rather than only summed \`bytes_transfer\` / \`bytes_body\`) so analysis can post-hoc filter concurrent fetches not attributable to the search.
- \`tests/test_search_perf.py\`:
  - replace deprecated \`dt.datetime.utcnow()\` with \`dt.datetime.now(UTC)\` (Python 3.12+ compat)
  - replace fixed 800ms \`wait_for_timeout\` after facet changes with \`wait_for_function\` polling on \`.facet-count.recomputing\` clearing — the cross-filter handler debounces 250ms then runs an async cluster reload that can exceed 1s cold; fixed waits flake on slow networks
  - add \`composed-source-material\` query to canonical set; exercises the \`facetFilterSQL()\` pid-IN-subquery path (uses \"first material checkbox\" rather than hard-coded URI to survive data-snapshot rotation)
  - tighten final assertion: refuse to treat a benchmark with any silent failures as valid (was \`>= half\`); partial JSON still on disk for diagnosis

**Deferred** (filed as #174):

- unit test for the instrumentation itself (\`terms_count\`, \`elapsed_ms\`, \`results_count\` invariants)
- formal definition of \"warm\" in the v1 contract budget table (#169)
- defensive wrap of \`searchTerms()\` once the v1 tokenizer lands (#170)
- tighter bytes attribution beyond the search window

rdhyee:
Review findings from this pass:

1. `?perf=1` panel will not show searches run after initial render. `explorer.qmd` only reads existing `performance.measure` entries while the OJS `perfPanel` cell renders once after `phase1`; later `doSearch()` calls log to console but never update the panel DOM. This undercuts the PR claim that search measures are appended to the panel per invocation. Fix by appending/updating the perf panel from `doSearch()` or wiring an explicit refresh path.

2. `_wait_for_facet_settle()` can return before the async facet/source work has even started. The helper waits for no `.recomputing`, but the app schedules recompute after a 250ms debounce and source changes call `refreshFacetCounts()` only after `await loadRes(...)` completes. The benchmark can start the search while cluster reload/facet recompute is still running, contaminating the composed-filter timings. The harness needs a positive “work started then finished” signal, or an explicit app-side settle marker.

3. `_measure_one_query()` leaks browser contexts on failures. Since the outer loop catches exceptions and continues, a timeout leaves a context/page alive and can skew later measurements. Wrap the body in `try/finally: context.close()`.

Verified only with `python -m py_compile tests/test_search_perf.py`; I did not run the full perf-smoke.

rdhyee:
## Codex review fixes (commit \`76e37b6\`)

All three findings addressed:

**Medium #1 (panel updates per search)**: \`doSearch\` finally block now appends a row to the \`?perf=1\` panel per search. The panel renders once at boot from existing measure entries; this hooks each subsequent search so the panel stays current. The PR claim "search measures are appended to the existing ?perf=1 panel per invocation" is now true.

**Medium #2 (busy-flag race)**: Source filter and material/context/object_type change handlers wrap their async work in try/finally with \`body.classList\` 'explorer-busy' set during the chain. The 300 ms post-\`refreshFacetCounts\` wait is intentional: \`refreshFacetCounts\` schedules a 250 ms debounce that THEN sets \`.recomputing\` on facet count spans; we hold busy until that fires so external observers see a coherent "work started → work finished" signal instead of racing the debounce. \`handleFacetFilterChange\` is now async and awaits \`loadViewportSamples\` in the point-mode branch (was fire-and-forget) so the busy flag honestly covers all triggered work.

\`_wait_for_facet_settle\` in the test is now two-phase: wait for \`body:not(.explorer-busy)\`, then wait for \`.facet-count:not(.recomputing)\`. Polling either signal alone races the app's two-stage settle.

**Low #3 (context leak on timeout)**: \`_measure_one_query\` body wrapped in try/finally on \`context.close()\`. A TimeoutError in \`_run_search\` no longer leaks the browser context.

**Open question on #176**: agreed — the spec MUST clause is not closed by doc honesty alone. Filed the UI-hint fix as commit \`006d335\` on PR #176.

Diff: +112/-43.

rdhyee:
## Round 2 fix (commit \`3b97e34\`)

You're right — the binary class racing on overlapping handlers was the bug behind the bug. Replaced with a depth counter:

\`\`\`js
let _busyDepth = 0;
function busyAcquire() {
    if (_busyDepth === 0) document.body.classList.add('explorer-busy');
    _busyDepth++;
}
function busyRelease() {
    _busyDepth = Math.max(0, _busyDepth - 1);
    if (_busyDepth === 0) document.body.classList.remove('explorer-busy');
}
\`\`\`

Class is added on the 0→1 transition and cleared only on the 1→0 transition. \`Math.max\`-guarded against double-release. Both handlers (source filter, material/context/object_type) now use \`busyAcquire\`/\`busyRelease\` in their try/finally blocks.

This means \`_apply_source_filter\` clicking 3 source checkboxes back-to-back now correctly holds the busy flag through the LAST handler's full work, not just the first/fastest.

Diff: +22/-4.

PR #177, opened 2026-05-08
explorer: interim recall fix — search description (#168 Direction A, knowingly slower)

rdhyee:
## Summary — interim recall fix, NOT the future FTS backend

\`doSearch()\` swapped from \`samples_map_lite.parquet\` to \`sample_facets_v2.parquet\` so live search now covers \`label + description + place_name\` instead of \`label + place_name\`. **This trades latency for honest recall**, knowingly violating the latency thresholds locked in #168.

The previous live search returned **zero** results for \`pottery Cyprus\` even though ~7,124 samples actually match (Cyprus appears only in description). The #168 honesty-fix landed in #176 documented this gap; this PR closes it functionally as an *interim* state while the proper substrate work in #169-#172 builds the real FTS that fixes both recall and latency.

## Why this is shipping despite the threshold failure

- #168 locked: cold \`pottery\` ≤ 5 s, multi-term ≤ 6 s.
- #167 baseline measured the swap at 12 s and 14.8 s respectively.
- **Strictly per the threshold rule, Direction B (the doc fix in #176) was the answer.** That's what merged.
- But the threshold was set when we expected the swap to be roughly latency-neutral. It isn't. And "fast and broken" (current behavior on \`pottery Cyprus\`: 5 s, returns 0 results) is worse UX than "slow and useful" (this PR: 15 s, returns 50 results).
- A search bar that tells you \`pottery Cyprus\` is a good example query, then returns nothing for it, is broken. We made the placeholder example honest first; this PR makes the underlying behavior honest.

This is the kind of decision the threshold rule was a *guess* about, not an answer to. The data flipped the answer.

## What this is NOT

- **Not the FTS solution.** #169-#172 builds the real substrate: sample-centric document projection (label + description + place_name + dereferenced concept labels), hash-partitioned BM25 indexes, sub-second cold search, sub-500-ms warm. This PR's interim search goes away the day #171 lands.
- **Not a reversal of the substrate decision.** The state contract (PR #166) and search-index contract (#169 / PR #175) stand. Track 6 (hosted-search) remains a contingency.
- **Not a precedent for relaxing locked thresholds.** Future locked thresholds remain locked unless similarly contested by data.

## What changes

[...]

Issue #172, opened 2026-05-08
Explorer FTS Track 5: GO/NO-GO decision gate


rdhyee:
> ## BLUF (plain English, updated 2026-07-17)
>
> **Where this stands:** the new fast search **shipped as the site default this morning**
> ([PR #335](https://github.com/isamplesorg/isamplesorg.github.io/pull/335); old search
> survives at `?fts=off`). That was a *relative* call — the new path beats the old on every
> measured row — **not** a pass of this gate: the §7 end-to-end budgets below measure
> search **plus** the page-wide refiltering that follows it, and that shared machinery
> now dominates for either backend.
>
> **The next step, and our position — just take it:** build the **pin-overlay UI**
> (option C, already adopted in #164/#165 — search fills the side panel and pins matches
> on the globe, instead of refiltering the whole page). It's the piece that makes search
> feel *instant* and makes this gate's budgets winnable as written. Est. days, not weeks.
>
> **The only reasons we'd hold off — say so if any of these bite:**
> 1. **Runway** — grant ends 7/31; pin-overlay competes with Zenodo closeout and the
>    ~1-day coherence-manifest v0 (#334). Our read: v0 + pin-overlay both fit if we start now.
> 2. **UX re-confirm** — option C changes behavior: search would *pin results on top*
>    rather than filter the facet counts/table. The decision is from May; if anyone's
>    expectation has drifted, now's the moment.
> 3. **SME ranking verdicts** ([A/B links](https://github.com/isamplesorg/isamplesorg.github.io/issues/172#issuecomment-5004367575))
>    could force ranking tuning first — but that work is independent and can run in parallel.
>
> Absent an objection on 1–3, pin-overlay implementation starts next. Historical gate
> spec follows below, unchanged.

---

> **Updated 2026-05-08** (rounds 1 + 2 per Codex review on #165). Round 1 added quality-gate cells. Round 2 sharpened: (a) "non-empty" is necessary but not sufficient — concept-only and stopword-heavy checks now require top-K relevance, not just any results; (b) added pathological-behavior hard-fails covering tokenizer parity, all-stopword queries, duplicate terms, edge-length tokens, missing display-join rows, filter composition; (c) NO-GO framing makes hosted-search a permanent contingency for *either* v1 failure or future v2+ quality requirements.

Sub-issue of #165. **Depends on #171 (browser query prototype + benchmark data).**

## Goal

[...]

PR #166, opened 2026-05-08
docs: explorer state contract (closes #164)

rdhyee:
## Summary

Adds `EXPLORER_STATE.md` — the state-contract doc for `explorer.qmd`, requested in #164.

- Inventory tables for every piece of explorer state: URL query params, URL hash params, DOM-as-state, `viewer.*` / `window.*` fields, OJS cells with deps + side effects.
- **Search-semantics decision: option C (side-panel lookup with result-pin overlay).** Updated from the original option-B framing after Codex review on #165 surfaced a sharper third option. The §6 implementation rules (50-pin cap, z-order above cluster + sample-mode, hollow-ring styling on source palette, fit-to-bounds **only when result extent < 30° × 30°**, lifecycle tied to `?search=` non-empty) and a result-set-shape acceptance table (zero / one / local-many / global-many) are non-negotiable defaults for the implementer.
- Facet-count contract restated to match #158 and the option-C decision (`?search=` does **NO** modify facet counts; backend contract unchanged from option B).
- All references cite `explorer.qmd` line numbers against current `main` (94e7674).
- Confirmed `_urlParamsHydrated` is gone.

PR is doc-only, no code changes. Three commits: original doc (`78ec90a`), option C addition (`d139313`), option C UX-rule + §9 reconciliation + pin-count wording (`cdab34a` + `062863d`).

## What this unblocks

Once merged, the items in #163 can be re-scoped against the contract:
- **#163 item 4** (zero search results + populated map looks broken) is *resolved* by option C, not by side-panel copy. Implementation must visually verify the four result-set shape cases (zero / one / local-many / global-many).
- **#163 item 6** (page in URL) gets a clear contract slot in the `writeQueryState` / hydration cycle.
- Items 1, 2, 3, 5, 7 are state-preserving UI fixes that don't perturb this contract.

## Coordinating with the parallel FTS workstream

The search-backend question is intentionally orthogonal to option C. The state contract specifies *which UI surfaces display matches* (side panel + result-pin overlay); the FTS substrate work in #165 / #169-#172 specifies *how matches are computed*. The two land independently; the contract is compatible with every backend the FTS work might choose (in-browser ILIKE, static-Parquet inverted index, hosted-search service).

## Test plan

- [ ] Review inventory tables against current `explorer.qmd`
- [ ] Confirm option-C decision and rationale (vs A and original B framing)
- [ ] Confirm option-C implementation rules table is non-negotiable defaults the implementer should respect
[...]

Issue #165, opened 2026-05-08
Improve Interactive Explorer full-text search substrate


rdhyee:
## Context

Follow-up to #84 and PR #95. The DuckDB FTS spike answered the first question: native DuckDB FTS works, including BM25 ranking, but the generated browser-delivered index is too large to ship by default.

As of 2026-05-08, the live Interactive Explorer at https://isamples.org/explorer.html does not have production full-text search in the search-engine sense. It has browser-side DuckDB-WASM substring search over Parquet.

## Current behavior

- The live page is backed by `isamplesorg.github.io/explorer.qmd`.
- Search runs against `https://data.isamples.org/isamples_202601_samples_map_lite.parquet` (~60 MB).
- The live search predicate covers only:
  - `label`
  - `place_name`
- Search uses `ILIKE '%term%'`, with multi-word queries split into terms and combined with `AND`.
- Result ordering is a simple score:
  - `label` match = 3
  - `place_name` match = 2
- Search composes with active source and facet filters.
- Search does not currently cover `description` in the live explorer because `samples_map_lite.parquet` does not carry it.
- Search does not cover the canonical Solr `searchText` equivalent described in `query-spec.qmd`.

Relevant code/docs:

- `explorer.qmd`: `searchTerms`, `textSearchWhere`, `textSearchScore`
- `explorer.qmd`: search handler around `doSearch()`
- `query-spec.qmd`: intended `text MATCHES` semantics
- `tools/build_fts_index.py`: preserved DuckDB FTS spike artifact

## Mismatch to resolve

`query-spec.qmd` currently says the web Explorer subset is `label + description + place_name`, but the live explorer searches only `label + place_name`.

Either:

1. Update the docs/UI to state the actual live subset, or
2. Change the live search implementation to include `description`, probably via `sample_facets_v2.parquet` or another search-optimized projection.

## Measured behavior

Native DuckDB measurements against the live hosted Parquet files, run on 2026-05-08. These are not browser timings, but they show the rough scan cost.

Live-style `samples_map_lite` search over `label + place_name`:

- `pottery`: 50 results, ~1.8s cold
- `basalt`: 50 results, ~0.5s
- unlikely no-hit term: 0 results, ~0.2s

`sample_facets_v2` search over `label + description + place_name`:

- `pottery`: 50 results, ~3.0s
- `cyprus`: 50 results, ~0.7s
- `basalt`: 50 results, ~0.5s

[...]

PR #179, opened 2026-05-08
explorer: two-button search semantics (closes #178, Light path)

### LINKED ISSUE #178: Interactive Explorer: integrate Hana mockup feedback (Light path on search semantics)
Captures the Interactive Explorer feedback from the 2026-05-08 tech call discussion of Hana's wireframe mockup ([Figma node 213:394](https://www.figma.com/design/Nqkuqh3Z4aqVh0nmwUAgKg/iSamples-Wireframe-1.0?node-id=213-394)). Hana noted "corrections coming"; this issue tracks the Light-path subset that integrates with our shipped FTS work without re-opening the locked contracts.

## Background

The mockup proposes a search UX that introduces a **third semantic** alongside our recent option C (locked in #164 / PR #166): two explicit buttons in the side panel — "Search Selected Areas" (viewport-scoped) and "Search Entire World" (full-corpus). EXPLORER_STATE.md §6 currently commits to option C with side-panel + result-pin overlay; the mockup overlays a viewport-scope toggle on top.

Two implementation paths:

- **Light** — keep option C as the default (full-corpus side-panel lookup with result-pin overlay); add "Search Selected Areas" as a viewport-scoped variant that runs the same query with a `latitude/longitude BETWEEN` predicate added. No contract change to #169-#172.
- **Heavy** — revisit option A vs C entirely as design intent shifts (full A/B/C/two-button rethink).

**Decision: Light first. Heavy deferred** until we've shipped the substrate work in #169-#172 and have data on whether viewport-scoped search materially shifts user behavior.

### UI

Two buttons in the side panel below the search input + facet checkboxes (matching the Figma mockup):

- **"Search Selected Areas"** (orange in mockup) — runs current query against samples whose lat/lng falls within the camera's current view rectangle.
- **"Search Entire World"** (blue in mockup) — runs the current full-corpus search (existing `doSearch()` behavior, post-PR #177).

The current single "Search" button → replaced by these two. Pressing Enter in the search input runs whichever was last clicked, defaulting to **Search Entire World** (preserves current behavior for keyboard users).

### Backend

In `doSearch(scope)`:

```js
async function doSearch(scope) {
    // ... existing instrumentation + token tokenization ...

    let viewportPredicate = '';
    if (scope === 'area') {
        const rect = viewer.camera.computeViewRectangle(viewer.scene.globe.ellipsoid);
        if (rect) {
            const south = Cesium.Math.toDegrees(rect.south);
            const north = Cesium.Math.toDegrees(rect.north);
            const west  = Cesium.Math.toDegrees(rect.west);
            const east  = Cesium.Math.toDegrees(rect.east);
            // Dateline-crossing case (west > east): split into two ranges.
            // The lat/lng columns live in the lite parquet, joined as `l` in the CTE.
            if (west > east) {
                viewportPredicate = `
                    AND l.latitude BETWEEN ${south} AND ${north}
                    AND (l.longitude BETWEEN ${west} AND 180
                      OR l.longitude BETWEEN -180 AND ${east})
                `;
            } else {
                viewportPredicate = `
                    AND l.latitude BETWEEN ${south} AND ${north}
                    AND l.longitude BETWEEN ${west} AND ${east}
                `;
            }
        }
    }

    const results = await db.query(`
        WITH matches AS (
            SELECT pid, label, source, place_name, (${score}) AS relevance_score
            FROM read_parquet('${facets_url}')
            WHERE ${searchWhere}
            ${sourceFilterSQL('source')}
            ${facetFilterSQL()}
            ORDER BY relevance_score DESC
            LIMIT 50
        )
        SELECT m.pid, m.label, m.source, l.latitude, l.longitude,
               m.place_name, m.relevance_score
        FROM matches m
        LEFT JOIN read_parquet('${lite_url}') l USING (pid)
        WHERE 1=1 ${viewportPredicate}
        ORDER BY m.relevance_score DESC, m.label
    `);
}
```

The viewport predicate goes on the **outer** query (post-join), not the CTE — because the inner CTE searches `sample_facets_v2` which doesn't have lat/lng. This means the CTE may return up to 50 matches that fail the viewport predicate; the outer LIMIT can drop us below 50 results in viewport-scoped mode. For v1 this is fine (users can pan to widen the area). A future tuning could increase the inner LIMIT to e.g. 200 when in area mode.

### URL state

- New query param `?search_scope=area|world`. Hydrated by `applyQueryToSearch()`; written by `writeQueryState()`.
- Default: `world` (omitted from URL).
- Camera position is the source of truth for what "area" means; no need to encode the rect in the URL — `?v=1&lat=…&lng=…&alt=…` already captures it.

### Result-pin overlay

The option C result-pin overlay (per EXPLORER_STATE.md §6) still applies for both scopes. Pin coordinates already reflect what was found; viewport-scope just narrows the candidate set.

### Open question

Does Hana's design intent want "Search Selected Areas" as the **primary** button (left position) or the secondary? Mockup shows orange (selected areas) on the left and blue (entire world) on the right, suggesting selected-areas is primary. If so, the keyboard default may want to flip to area-scope. Defer until confirmed with Hana/Andrea.

## Other mockup-derived items (separate from search semantics)

These are net-new UX work, no conflict with shipped contracts. Not in scope for the Light search-semantics PR; tracked here for future follow-ups:

- **Educational tooltips** ("What is a Cluster?", "What is a Sample?") — onboarding boxes on the left edge of the explorer.
- **Tree-selection vocabularies** for material / sample / sample-feature hierarchies — Hana to make a mockup diagram.
- **Top-level sample-type icons** (5 per category) for missing-image fallback in sample cards.
- **Visual accessibility**: halos around dots for color-blind / low-vision users.
- **Native Cesium control panel** on the left edge — currently hidden; mockup shows it visible.
- **Quick Stats panel restructure** — Selected Cluster moved into the same table as Quick Stats.
- **Visible "Copy Link to Selected View" button** — currently we have a `shareBtn`; mockup styles it more prominently with a labeled button.
- **Permanent sample-rows table at the bottom** — mockup shows the table view always-visible under the globe rather than as a Globe/Table toggle. May conflict with current binary view (#162); needs a UX call.

Each of these gets its own issue once Hana's "corrections coming" iteration lands.

## Acceptance for this issue (Light path only)

- [ ] Two buttons in the side panel: "Search Selected Areas" + "Search Entire World"
- [ ] `?search_scope=area|world` URL param, hydrated + written
- [ ] `doSearch(scope)` SQL conditionally adds the viewport predicate, with dateline-crossing handled
- [ ] Result-pin overlay continues to work in both modes
- [ ] Perf-smoke (#173) extended with one viewport-scoped query case in the canonical set
- [ ] EXPLORER_STATE.md §6 amended: option C remains the default; the area-scoped variant is documented as a Light extension that does not require revisiting A/B/C

## Out of scope

- Heavy revisit of EXPLORER_STATE.md §6 (full A/B/C/two-button rethink).
- The other mockup-derived items listed above.
- Any backend change to the substrate work in #169-#172.

## Refs

#163 (UX rework umbrella), #165 (FTS umbrella), PR #166 (state contract option C), PR #177 (interim recall fix), Hana mockup [Figma 213:394](https://www.figma.com/design/Nqkuqh3Z4aqVh0nmwUAgKg/iSamples-Wireframe-1.0?node-id=213-394).

rdhyee:
## Summary

Light path of #178 — Hana's mockup feedback for the Interactive Explorer (discussed in 2026-05-08 tech call). Replaces the single "Search" button with two scope buttons:

- **Search Selected Areas** (orange) — limits results to samples whose lat/lng falls in the current camera view rectangle.
- **Search Entire World** (blue) — runs the existing full-corpus search.

Backend stays option C (per [EXPLORER_STATE.md §6](https://github.com/isamplesorg/isamplesorg.github.io/blob/main/EXPLORER_STATE.md)) — search is still side-panel + result-pin overlay; cluster layer and facet counts are unaffected. Heavy revisit (full A/B/C rethink) explicitly deferred per #178.

## Verified locally

Manual browser eyeball check at \`http://localhost:5880/explorer.html?perf=1#v=1&lat=35&lng=33&alt=2000000\` (Cyprus camera position):

| step | result |
|---|---|
| Page loads, two buttons visible, old \`#searchBtn\` removed | ✓ |
| Type \`pottery\` → click "Search Selected Areas" | URL becomes \`?perf=1&search=pottery&search_scope=area\`, search runs |
| Area-scoped result | **0 results** — verified natively that none of the top-50 \`pottery\` matches fall in lat 30-40 / lng 25-40 |
| Camera stays at Cyprus (auto-fly suppressed) | ✓ |
| Click "Search Entire World" | URL drops \`search_scope\` param; **50+ results**; camera flies to top-1 (Italy) |
| Perf panel row shows \`search #N area: "pottery" (0) 4.02 s\` | ✓ |

The 0-result outcome from Cyprus is correct behavior per the contract — the area is genuinely too tight to contain any of the top-50 hits. Users widen by panning out, which the search-help and the second button signal.

## Implementation notes

**SQL shape**: viewport predicate goes on the OUTER query (post-join), not the inner CTE, because lat/lng live in \`samples_map_lite\` not \`sample_facets_v2\`. This means area-scoped searches can return < 50 results when the inner top-50 don't all satisfy viewport. Acceptable v1 behavior. A future tuning could increase the inner LIMIT in area mode if data shows users hit this bound often.

**Dateline-crossing**: when \`west > east\` (camera over the international date line), the longitude predicate splits into two ranges:

\`\`\`sql
AND (l.longitude BETWEEN ${west} AND 180
  OR l.longitude BETWEEN -180 AND ${east})
\`\`\`

**URL persistence**: \`persistSearchScope()\` is separate from \`writeQueryState()\` because the latter doesn't know about scope. \`?search_scope=area\` is written when area is chosen; deleted when world (default).

**Auto-fly suppression**: only \`world\` searches fly to the first result. \`area\` searches preserve the user's current camera — they're already where they want to be.

**Enter key**: uses the last-clicked scope, or URL-hydrated scope on cold boot. Defaults to \`world\` for keyboard-only first-time users.

## Doc + tests

- [EXPLORER_STATE.md §6](https://github.com/isamplesorg/isamplesorg.github.io/blob/main/EXPLORER_STATE.md): new "Light-path addendum" explains the two-button design as an extension of option C; Heavy revisit deferred until #170-#172 land.
- \`tests/test_search_perf.py\`: new \`area-scope\` canonical query uses \`url_hash\` to set the camera before clicking \`#searchAreaBtn\`. \`_run_search\` takes a \`scope\` param routing to the right button.

## Out of scope (other Hana mockup items, deferred)

Listed in #178: educational tooltips, vocabulary tree-selection, sample-type icons, accessibility halos, native Cesium control panel, table-always-visible. Each gets its own issue once Hana's "corrections coming" iteration lands.

## Test plan

- [ ] Open Explorer, type a query, click both buttons. Verify URL updates correctly.
- [ ] At a global camera view, both buttons should return ~50 results.
- [ ] Zoomed in tightly, "Search Selected Areas" should return < 50 (or 0) depending on what's in view.
- [ ] \`?perf=1\` panel rows include scope label.
- [ ] Reload with \`?search=foo&search_scope=area\` — search runs in area mode on boot.

Closes #178. Refs #163, #165, #166, #177.


rdhyee:
Review finding:

**Medium:** `Search Selected Areas` can return false zeroes because the viewport filter runs after the global top-50 limit. In `explorer.qmd`, the CTE ranks and limits 50 matches from `sample_facets_v2` first, then the outer query joins coordinates and applies `viewportPredicate`. That means area mode only searches “current viewport among the global top 50,” not “top 50 within the current viewport.” For broad terms like `pottery`, this can show `0 results` in a real area that has matches, simply because none of that area’s samples made the global top 50. This undercuts the button’s promise and reintroduces the false-zero problem the interim search work is trying to reduce.

The fix is to make area scope apply before the top-K selection. Since lat/lng live in `samples_map_lite`, area mode likely needs a separate query shape that joins `lite` inside the candidate CTE before `ORDER BY ... LIMIT 50`. If that is too slow, use an explicit larger candidate cap and document it as approximate, but the current exact-looking label is misleading.

Test gap: the perf harness adds an `area-scope` canonical query, but the committed `tests/search_baseline_2026-05-08.json` still has only the prior 9 labels. If the baseline artifact is meant to track the canonical query set, it should be rerun or explicitly left as historical pre-#179 data.

I did not run the full browser perf-smoke.

rdhyee:
## Round-2 fix (commit \`cc79ec0\`)

Codex was right — the previous shape had a real semantic bug. The viewport predicate ran AFTER the global top-50 selection, so "Search Selected Areas" was actually "current viewport among the global top 50," not "top 50 within the current viewport."

The bug was visible-in-the-wild: for \`pottery\` the global top-50 is dominated by one Alaska collection (label='Pottery AM662:...', all at lat=57.7 lng=-152.4). Any non-Alaska area-scope query for \`pottery\` would return zero — including the canonical Cyprus case the interim search work was built to fix.

### Fix

Split into two SQL shapes (committed):

| mode  | shape                                                                                                                                                              |
|-------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| world | CTE over sample_facets_v2 → top-50 → LEFT JOIN samples_map_lite. Unchanged. Coord-less samples still appear in results.                                            |
| area  | INNER JOIN samples_map_lite inside the candidate selection. Viewport BETWEEN applied BEFORE \`ORDER BY ... LIMIT 50\`. Drops coord-less samples (required by area). |

Both use \`f.\`-qualified column names so the same \`searchWhere\`/\`score\` strings work for both.

### Verified

**Native** (TIGHT Cyprus rect lat 30-40 lng 25-40):

| SQL form | top-50 result |
|---|---|
| Old (filter after CTE LIMIT) | **0** ← false-zero bug |
| New (filter inside CTE) | **50** ✓ |

**Browser** at Cyprus camera (\`lat=35, lng=33, alt=1Mm\`): "Search Selected Areas" for \`pottery\` → 50+ results, all clustered at the Dead Sea pottery site (31.13, 35.53). Exactly what the user expects.

### Doc

EXPLORER_STATE.md §6 Light-path addendum updated to describe both SQL shapes and why area mode drops coord-less samples.

Baseline JSON refresh coming in a follow-up commit on this branch — the perf-smoke is running now against the fixed code; the previous \`tests/search_baseline_2026-05-08.json\` was from a build before the area-scope canonical query was added (and before this fix).

Diff: +89 / -54.

rdhyee:
## Refreshed baseline (commit \`d37f703\`)

Perf-smoke against the fixed code, all 10 canonical queries:

| label                         | results | cold (ms) | warm (ms) |
|-------------------------------|--------:|----------:|----------:|
| single-common                 |      50 |    10534  |     4566  |
| single-rare                   |      50 |     9744  |     4493  |
| multi-term                    |      50 |    10013  |     4515  |
| no-hit                        |       0 |     9840  |     4644  |
| wildcard-pct                  |      50 |    12216  |     4362  |
| wildcard-under                |       0 |    10134  |     4575  |
| diacritic                     |      50 |    13237  |     4878  |
| composed-source               |      50 |     6072  |     5486  |
| composed-source-material      |       0 |     6582  |     3427  |
| **area-scope**                |  **50** | **10483** |  **4165** |

\`area-scope\` now returns 50 results (previously 0 with the pre-fix SQL). Latency envelope ~10-13s cold, ~4-5s warm — same shape as world mode; no regression from adding the INNER JOIN inside the area-mode CTE.

Also fixed: the \`field_subset\` JSON metadata string (was stale: \`"label+place_name samples_map_lite"\` from before #177 Direction A; now reads \`"label+description+place_name (sample_facets_v2 + lite for coords; world via LEFT JOIN, area via INNER JOIN with viewport predicate inside CTE)"\`).

Issue #163, opened 2026-05-01
Interactive Explorer rethink: architecture review + UX/feature backlog


rdhyee:
Follow-up to #156 / PR #162 (Phase 5 unified explorer). Captured during post-merge review on rdhyee staging at https://rdhyee.github.io/isamplesorg.github.io/explorer.html. Filing for design + scoping rather than as immediate-blocker tasks.

## 1. Number-of-samples discrepancy in hero copy

**Observation (Raymond)**: page subtitle says "Search and explore 6.7 million material samples" but the stats panel reports e.g. \`5,980,282 samples in view\`. The 6.7M is the dataset total; the lower number is samples with coordinates (likely — needs verification).

**Reaction (CC)**: yes, the gap is georeferenced rows. The per-source facet counts compound the confusion — a user sees SESAR=4.6M in the legend, then a smaller in-view count, with no explanation. Two non-exclusive fixes:
- Hero copy: \`6.7M samples · 6.0M with coordinates\` (or similar — confirm exact number against the lite parquet)
- Separate \"dataset total\" from \"on the map\" in the side panel so the relationship is legible

## 2. Tags on the page

**Current**: \`PARQUET\`, \`SPATIAL\`, \`H3\`, \`PERFORMANCE\`, \`ISAMPLES\` rendered as a tag row near the title.

**Observation (Raymond)**: rethink — these read like internal repo categorization, not user-facing.

**Reaction (CC)**: agree. Options: (a) drop the tag bar entirely — this is the marquee page, doesn't need self-categorization; (b) replace with user-facing tags like \`interactive\`, \`globe\`, \`search\`, \`samples\`. Lean toward (a).

## 3. H3 res4 / res6 / res8 labels

**Current**: the \"How This Demo Works\" table uses literal H3 resolution numbers and the stats panel shows \`H3 Res4\` etc.

**Observation (Raymond)**: technically accurate but maybe too techie.

**Reaction (CC)**: agree. Labels could be \`Continents (instant)\` → \`Regions\` → \`Neighborhoods\` → \`Individual samples\`. Keep the \"instant / 5K / 1.4M / 90M\" perf story in the deep-dive table but don't front-load H3 jargon.

## 4. Search results vs map sync

**Observation (Raymond)**: at https://rdhyee.github.io/isamplesorg.github.io/explorer.html?search=pottery+Cyprus the search panel says \"no results\" but all the dots stay on the map. Expected: empty map.

[...]

PR #182, opened 2026-05-09
docs: proposal — URL-persist selected cluster identity

rdhyee:
## Summary

Doc-only proposal. Companion to [`EXPLORER_STATE.md`](./EXPLORER_STATE.md). Audits what selection state the Explorer URL captures today and proposes adding **cluster** selection as a single `&h3=<cell>` hash token. No code changes.

**Goal**: enable URLs that replay "I clicked this specific dot" — both for sample dots (already works via `&pid=`) and **cluster dots** (currently in-memory only). The use case is debugging conversations: "tell me about this dot, why does it disappear when I zoom to res 6?".

## Audit findings

| Selection type | URL-persisted today? | Mechanism |
|---|---|---|
| **Sample** (single PID) | ✅ Yes | `#pid=<urlencoded>` — round-trips through `readHash` → `_globeState.selectedPid` → side-panel sample card + lazy `wide_url` description fetch (`explorer.qmd:892, 1813, 2167`) |
| **Cluster** (H3 cell aggregation) | ❌ No | Cluster-click *clears* `selectedPid` (`:919`) and writes nothing else; the cluster card + nearby-samples list are DOM-only and lost on reload |

## Recommended encoding (revised after Codex review)

**Option C: `&h3=<cell>` — H3 cell index**

```
#h3=841a067ffffffff
```

Codex review of v1 confirmed `h3_cell` is already SELECTed at `explorer.qmd:973` (phase1) and `:1316` (loadRes); only the cluster `.id` object needs to carry it forward. So:

- **Exact lookup**: `WHERE h3_cell = ?` is a primary-key join. The earlier Option B (lat/lng-tuple) recommendation was lossy and is rejected.
- **Single token, 15 hex chars**, resolution implicit in the cell.
- **No data-pipeline change** — purely a runtime fix to thread `h3_cell` through.
- **`?sources=` already URL-persisted** — the only filter that affects cluster aggregation (per `:1706-1710`).
- **No backwards-compat tax** — `&pid=` URLs keep working unchanged; mutual exclusion preserved at runtime.

## Scope of cluster reproducibility

Cluster identity depends on:
- The H3 cell (encoded in `&h3=`)
- The `?sources=` filter (already URL-persisted)

It does **not** depend on `?material=` / `?context=` / `?object_type=` — H3 summary parquets only carry `dominant_source`, and the code at `:1706-1710` explicitly documents that those facet filters cannot affect cluster counts.

## Open questions for review

1. **Cross-resolution behavior on URL load**: should `&h3=...&alt=...` *force* the camera/altitude to match the cell's natural resolution, or just populate the side panel? Recommendation in the doc: populate the side panel only.
2. **Does the side panel need a "stale cluster" hint** if the recipient's filters cause the on-globe rendering to differ from the URL one? Defer until we see whether it's confusing in practice.

## Phasing

- **Phase 1** *(this proposal)*: ship `&h3=` per Option C. ~25-line patch in `explorer.qmd`. Update `EXPLORER_STATE.md §2` to add the row.
- **Phase 2** *(only if needed)*: unified `&sel=` field for a hypothetical third selection type (e.g., region/polygon).

The previously-planned Phase 2 compat shim (`&cluster=` → `&h3=`) is dropped — going straight to the canonical encoding means no lossy intermediate form to migrate from.

## Review history

- `ffc744d` — initial v1 doc (recommended Option B, treated H3 cell index as an open question).
- `877afcd` — revised per Codex review: switched to Option C, narrowed filter-dependence section to source-filter only, simplified phasing.

Refs `EXPLORER_STATE.md`, #163 (UX rework umbrella).

### DIFF
--- a/EXPLORER_CLUSTER_URL_PROPOSAL.md
+++ b/EXPLORER_CLUSTER_URL_PROPOSAL.md
@@ -1,0 +1,193 @@
+# Explorer Selection URL State — Audit + Cluster Proposal
+
+Companion to [`EXPLORER_STATE.md`](./EXPLORER_STATE.md). Audit of what selection state
+the Explorer URL captures today, and a proposal for adding cluster-selection
+state so a URL alone can replay "I clicked this dot and got these samples."
+
+All file:line references are against `explorer.qmd` at commit `e0043c8`
+(post-#180 polish PR).
+
+---
+
+## 1. Audit: what's URL-persisted vs in-memory only
+
+### Already URL-persisted
+
+These already round-trip through the URL today; no work needed.
+
+| State | URL token | Notes |
+|---|---|---|
+| Camera position | `#lat`, `#lng`, `#alt`, `#heading`, `#pitch` | Hash; debounced 600 ms; `EXPLORER_STATE.md §2`. |
+| Mode (cluster vs point) | `#mode=point` (absent ⇒ cluster) | Hash. Pushed on `enter/exitPointMode`. |
+| **Selected sample** | **`#pid=<urlencoded>`** | Hash. `selectedPid` written by sample-click (`:892`), cleared by cluster-click (`:919`). Round-trips through `readHash` → `_globeState.selectedPid` → `updateSampleCard()` + lazy `wide_url` description fetch (`:1813`, `:2167`). |
+| Search query | `?search=` | Query-string. `EXPLORER_STATE.md §1`. |
+| Search scope | `?search_scope=area\|world` | Query-string. From PR #179. |
+| View mode (globe/table) | `?view=table` (absent ⇒ globe) | Query-string. |
+| Source filter (4 source toggles) | `?sources=CSV` | Query-string. |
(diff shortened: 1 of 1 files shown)


rdhyee:
Pushed revision `877afcd` integrating Codex's two P2 findings:

1. **H3 cell index is not unknown** — already in the parquet and SELECTed at `explorer.qmd:973` and `:1316`. Switched recommendation from Option B (`&cluster=SESAR:4:lat,lng`, lossy tuple) to **Option C (`&h3=<cell>`, exact key join)**. No data-pipeline change required; only thread `h3_cell` into the runtime cluster `id` object.

2. **Filter dependence was overstated** — H3 summary parquets only carry `dominant_source`; the code at `:1706-1710` documents that material/context/object_type filters cannot affect cluster counts. Tightened §2 to reflect this; cluster reproducibility only depends on `?sources=` which is already URL-persisted.

Phasing simplified: Phase 1 now ships Option C directly. The previously-planned Phase 2 compat shim (migrating `&cluster=` → `&h3=`) is dropped — by going straight to the canonical encoding we never ship the intermediate lossy form.

Doc-only PR. Ready for re-review whenever convenient.

PR #186, opened 2026-05-09
explorer: persist selected cluster identity in URL via &h3=<cell> (Phase 1)

rdhyee:
## Summary

Phase 1 of [`EXPLORER_CLUSTER_URL_PROPOSAL.md`](./EXPLORER_CLUSTER_URL_PROPOSAL.md) (merged in #182). Cluster selection now round-trips through the URL hash, complementing the existing `&pid=` for samples.

**Use case**: share or bookmark a specific cluster you clicked, have collaborators land on the same H3 cell with the side-panel populated. Specifically aimed at debugging conversations: "tell me about this dot, why does it disappear when I zoom to res 6?".

## Encoding

```
#h3=843f6d3ffffffff
```

H3 cell index in canonical 15-char hex (no `0x` prefix). The cell index encodes its own resolution; no separate `&res=` or `&cluster_source=` field needed. `&sources=` (already URL-persisted) covers the only filter that affects cluster aggregation — material/context/object_type filters can't, per the comment at `:1706-1710`.

## Mechanics

| Concern | Implementation |
|---|---|
| Cluster id carries h3_cell | `row.h3_cell.toString(16)` at both `add()` sites (`:992`, `:1335`). Parquet stores UBIGINT; we convert to hex once at ingestion. |
| `_globeState.selectedH3` | Mutated by cluster-click (`:923`); cleared by sample-click (`:895`) for mutual exclusion. Mirrors `selectedPid`. |
| URL emit | `buildHash` writes `&h3=` when `selectedH3` non-null (`:645`). |
| URL parse | `readHash` parses `params.get('h3')` (`:626`). |
| Cross-resolution lookup | `fetchClusterByH3` (`:1791`) UNIONs across res4/res6/res8 with `WHERE h3_cell = CAST('${decimal}' AS UBIGINT)`. DuckDB-WASM doesn't accept `0x...` integer literals, so hex → decimal conversion happens in JS via BigInt. |
| Hydration UX | `hydrateClusterUI` (`:1827`) mirrors the cluster-click side-panel + nearby-samples query. Called from both boot deep-link (`:2266`) and `hashchange` (`:1899`). |
| Mutual exclusion at hydration | `&pid=` wins if both `&pid=` and `&h3=` are present, per proposal §4. |

`EXPLORER_STATE.md §2` updated with the new `h3` row.

## Verified locally

`quarto render` + Playwright. Test URL (a known res4 cell with 151,334 OpenContext samples in central Turkey):

```
http://localhost:5880/explorer.html#v=1&lat=37.66&lng=32.83&alt=2000000&h3=843f6d3ffffffff
```

Result:

| check | result |
|---|---|
| Side panel shows `Selected Cluster` card | ✓ "OpenContext / H3 res4 / 151,334 samples / 37.6619, 32.8334" |
| Nearby samples list populates | ✓ 30 OpenContext radiocarbon samples |
| URL without `h3`/`pid` | ✓ loads cleanly |
| Hex → decimal round-trip via BigInt | ✓ 595590966136537087 looks up correctly |

The cluster-click → URL emission path uses the same `meta.h3_cell` field that `add()` populates; verified by static review (couldn't reliably synthesize Cesium picks via Playwright in this run).

## Test plan

- [ ] Visit `https://isamples.org/explorer.html#v=1&lat=37.66&lng=32.83&alt=2000000&h3=843f6d3ffffffff` — side panel populates without clicking anything
- [ ] Click a cluster — URL gains `&h3=<cell>`; reload → side panel re-populates
- [ ] Click a sample — URL has `&pid=` and *not* `&h3=` (mutual exclusion)
- [ ] Click another cluster after a sample — `&pid=` clears, `&h3=` set
- [ ] Browser back/forward through cluster/sample selections — side panel updates accordingly
- [ ] An unknown `&h3=` value (cell not in any parquet) — page loads cleanly without errors

Refs #182 (proposal), `EXPLORER_STATE.md`, #163.


rdhyee:
Pushed `a6c328f` addressing Codex review:

| Codex finding | Fix |
|---|---|
| **(a) Blocker** — async race in hashchange (stale fetch repainting newer hash) | Monotonic `viewer._selGen` bumped per hashchange, checked after every await |
| Strict regex over silent strip | `/^[0-9a-f]{15}$/i.test(lower)` |
| Normalize to canonical lowercase from lookup result | `selectedH3 = meta.h3_cell` after fetch, both in hashchange and boot deep-link |
| `else if` in `buildHash` for pid/h3 mutual exclusion | Done |
| UNION ALL → route by 2nd hex char | `RES_TO_H3_URL[parseInt(lower[1], 16)]` — one fetch instead of three |
| Unknown h3 leaves stale side-panel content | Now actively clears cluster card + samples list |

Verified locally:
- Uppercase `#h3=843F6D3FFFFFFFF` → hydrates correctly; runtime canonicalizes for subsequent `buildHash` writes
- Unknown well-formed cell → side panel cleared cleanly
- Non-hex `#h3=zzz_NOT_HEX_zzz` → silently rejected, no JS errors
- Known-good cell → round-trips unchanged

Ready for re-review.

rdhyee:
Pushed `1e8ef15` addressing Codex's second-round blocker:

| Finding | Fix |
|---|---|
| **Blocker (residual race)** — `hydrateClusterUI`'s internal `await db.query` for nearby samples could still call `updateSamples` for a stale h3 even with the outer-level `selGen` guard | Threaded an optional `isStale` predicate into `hydrateClusterUI`; checked after the inner await before `updateSamples` and the catch-path's error message. Hashchange caller passes the predicate; click and boot callers leave it undefined (no race possible there). |
| **Nit** — bump `_selGen` *before* the lat/lng early return | Done — bump moved to top of hashchange handler |
| **Nit** — defensive guard against non-cell H3 modes | Added `if (lower[0] !== '8') return null;` in `fetchClusterByH3` |

Known-good `#h3=843f6d3ffffffff` still round-trips locally.

Ready for re-review.

rdhyee (commented):
Findings:

- P2: Boot deep-link hydration can still race with a later hashchange. explorer.qmd around the boot deep-link path still calls fetchClusterByH3(ih.h3) and then hydrateClusterUI(meta) without the freshness predicate. The hashchange listener is registered earlier, so a slow initial &h3= lookup can be superseded by browser back/forward or a manual hash change, then the boot path can finish later and repaint the side panel with the original stale cluster. The same _selGen guard should cover the boot path too, or boot should abort if location.hash no longer matches the initial ih.h3.

- P2: fetchClusterByH3 bypasses the active source filter. The lookup query reads WHERE h3_cell = ... but does not include sourceFilterSQL("dominant_source"). Since this PR contract says ?sources= is the one cluster-affecting filter, an &h3= URL whose active sources excludes that cluster can still hydrate a cluster card for a dot that is not visible under the current filter state. That can also produce a mismatched panel: unfiltered cluster card/count/source, but nearby samples filtered by sourceFilterSQL("source") inside hydrateClusterUI.

The hydrateClusterUI internal freshness check from comment 4413059240 fixes the second-round race I had flagged inside the nearby-samples query, but the boot path and source-filter consistency still need tightening.

rdhyee:
Pushed \`90d3fdf\` addressing Codex's v3 findings:

| Finding | Fix |
|---|---|
| **P2** — boot deep-link racing with later hashchange | Boot path now bumps \`_selGen\` and uses an \`isBootStale\` predicate, threaded through all four awaits (pid lookup, wide-parquet description fetch, h3 lookup, hydrateClusterUI). |
| **P2** — \`fetchClusterByH3\` ignoring \`?sources=\` filter | Added \`sourceFilterSQL('dominant_source')\` to the cluster lookup. Excluded sources → returns null → side panel stays empty (matches the globe). No more mismatched cluster-card-vs-samples-list. |

Verified locally:
- \`?sources=SESAR,GEOME,SMITHSONIAN#h3=843f6d3ffffffff\` (cluster's source OPENCONTEXT excluded) → side panel correctly empty.
- \`?sources=\` defaults #h3=843f6d3ffffffff → hydrates with 151,334 samples / 30 nearby rows as before.

Ready for fourth-round review.

rdhyee:
Pushed \`ebd7978\` addressing Codex's v4 findings:

| Finding | Fix |
|---|---|
| Source-filter changes don't invalidate selection state | Source-filter handler now bumps \`_selGen\` and re-validates selection under the new filter — cluster via \`fetchClusterByH3\` (honors source filter), sample via lite_url probe with sourceFilterSQL. Filtered-out → clear selection + side panel + URL. |
| Boot's stale-abort early-returns skipped \`_suppressHashWrite = false\` | Wrapped boot deep-link block in try/finally; \`_suppressHashWrite = false\` moved to the finally so it always runs. |

Verified locally: load \`#h3=843f6d3ffffffff\` (OpenContext cluster) → side panel hydrates. Uncheck OPENCONTEXT in legend → \`&h3=\` drops from URL, cluster card empties, samples list clears, \`?sources=\` rewrites with the other 3 sources.

Ready for fifth-round review.

rdhyee:
Pushed \`0ee44b7\` addressing Codex v5 P2 + a latent precision-loss bug it surfaced.

**Codex v5 P2: cluster surviving filter wasn't being rehydrated.**
When the user unchecked a non-cluster source (e.g. SESAR while the selected cluster's source is OPENCONTEXT), the cluster card stayed but the nearby-samples list could show stale rows from unchecked sources or miss newly-checked ones. Fix: in the source-filter handler's revalidate branch, when \`meta\` is truthy, call \`hydrateClusterUI(meta, isStale)\` — not just leave the panel.

**Latent UBIGINT precision-loss bug (surfaced by testing #1).**
DuckDB-WASM returns \`h3_cell\` (UBIGINT >> 2^53) as a JS Number. \`row.h3_cell.toString(16)\` was therefore lossy: \`843f6d3ffffffff\` got rounded to \`843f6d400000000\`. Boot superficially worked because the SQL \`WHERE\` matched at the parquet level, but the lossy hex got stored in \`selectedH3\`; subsequent revalidations of the corrupted key never matched. Latent in \`ebd7978\`; the new rehydrate branch made it visible.

Fix: SQL \`SELECT CAST(h3_cell AS VARCHAR)\` returns a decimal string; JS converts via \`BigInt(decString).toString(16)\` — no precision loss. Applied at the two cluster-render sites (\`phase1\`, \`loadRes\`); \`fetchClusterByH3\`'s return now uses the validated input \`lower\` as the canonical hex. (Tried \`to_hex()\` first; DuckDB-WASM returns "Catalog Error: Scalar Function with name to_hex does not exist!".)

Verified locally:
| step | result |
|---|---|
| Boot at \`#h3=843f6d3ffffffff\` | cluster card populated, 30 samples, h3 in URL |
| Uncheck SESAR (cluster's OpenContext stays) | cluster card unchanged, samples re-rendered with only OpenContext rows, \`&h3=\` preserved |
| Uncheck OPENCONTEXT (cluster's own source) | card + samples cleared, \`&h3=\` dropped |

Ready for sixth-round review.

Issue #185, opened 2026-05-09
explorer: halo crescents / disappearing dots over hilly terrain (root cause not yet identified)


rdhyee:
## Symptom

After PR #180 added 1.5px white halos to point primitives in the Interactive Explorer, dots over hilly/mountainous terrain render with **partial halos (crescents)** at moderate zoom, and **disappear entirely** at close zoom over higher terrain.

**Reproducer**: https://isamples.org/explorer.html#v=1&lat=33.2706&lng=-86.2375&alt=311435&heading=360.0 — Birmingham, AL hill country.

Visible in the deployed prod site (current HEAD `9d9291c`, post-revert).

## What we've ruled out

Two fix attempts so far, both reverted/closed:

- **PR #181** (`d681758`, reverted via #183): added `disableDepthTestDistance: Number.POSITIVE_INFINITY` to all three `PointPrimitive.add()` sites. Caused **worse** regressions (Codex review): back-side-of-globe primitive bleed-through and broken hover/click pickability. Cesium docs confirm `POSITIVE_INFINITY` disables depth-test against the globe ellipsoid, not just terrain.
- **PR #184** (closed as no-op): set `scene.globe.depthTestAgainstTerrain = false` at viewer init. Codex caught this — it's already the Cesium 1.127 default. The artifact occurs *despite* depthTestAgainstTerrain being false.

So whatever's causing the crescents, it's **not the standard Cesium terrain depth-test**.

## Hypotheses worth checking

- **Alpha-blended fill becoming invisible**: `withAlpha(0.8)` / `0.85` / `0.9` on the fill (`color`) means at small pixel sizes the fill blends into bright satellite imagery, leaving only the opaque white outline → reads as an empty ring or crescent.
- **Cesium PointPrimitive shader artifact** at certain `pixelSize / outlineWidth` ratios under `scaleByDistance` (NearFarScalar).
- **Atmosphere or translucent ocean layer** occluding from one direction at low altitudes.
- **GPU / DPR-specific**: I cannot reproduce locally at canvas ~919×689 / DPR=1; the user's repro screenshot is at much higher resolution. Possibly a high-DPR-only artifact.

## What was attempted

| PR | Attempt | Outcome |
|---|---|---|
| #181 | `disableDepthTestDistance: Infinity` on all primitives | Reverted (#183) — bigger regressions |
| #184 | `globe.depthTestAgainstTerrain = false` | Closed — no-op, matches Cesium default |

[...]

PR #188, opened 2026-05-09
explorer: extract freshSelectionToken() primitive (closes #187 step 1)

### LINKED ISSUE #187: explorer: extract _selGen freshness primitive (post-mortem on PR #186 6-round Codex loop)
## Background

PR #186 (cluster URL state, `&h3=<cell>`) went through six rounds of Codex review. Every round surfaced at least one P2 issue, all of them resolved before merge. This issue is a post-mortem: was that iteration count a sign of architectural rot, scope creep, or genuinely complicated interactions — and is there a small refactor that would have prevented most of it?

Filed at the author's request after a meta-conversation with Claude Code; will post Codex's independent take as a follow-up comment, then we discuss.

## Author's question (verbatim)

> are the 6 rounds of iterations needed to converge a sign of the need for some architectural refactoring, creeping complexity, or there are just a lot of darn complicated interactions that need trial and error to surface?

## Claude Code's analysis

**Honest read**: ~80% of the findings are the same class of bug — async ops mutating shared state where the state can change mid-await — surfaced in different code paths. The remaining ~20% is a data-layer surprise (UBIGINT precision) plus a doc lag.

Looking at it that way:

| Round | Surface | Class |
|---|---|---|
| R1 | hashchange handler | async + shared-state race |
| R2 | hydrateClusterUI internal await | same |
| R3a | boot deep-link | same |
| R3b | fetchClusterByH3 ignoring source filter | data consistency |
| R4a | source-filter doesn't invalidate selection | same |
| R4b | suppressHashWrite leak on early-return | cleanup edge |
| R5a | rehydrate-on-filter when meta truthy | same data consistency |
| R5b | UBIGINT > 2^53 → Number precision-loss | data-layer surprise |
| R6 | stale docs | nit |

So this was mostly **one problem reappearing in different surfaces**, not creeping complexity from feature pile-up.

### The actual architectural smell

There's no single "selection store." Every path that touches selection (click handler, hashchange, boot, source-filter handler) imperatively does its own DOM writes, URL writes, and `_globeState` mutations. The `_selGen` token is a poor person's race guard duct-taped onto each path. We're now using it in four places.

### A small refactor would prevent future bugs of this class

1. **`withFreshness(token, fn)`** — a wrapper that auto-bails the inner function's DOM mutations if the token's generation has bumped. ~15 lines.
2. **(Optional, larger)** consolidate selection mutations into a `selectSample(pid)` / `selectCluster(h3)` / `clearSelection()` trio that handles state + URL + DOM in one place.

Step 1 alone would have prevented R2 + R5a (the "internal await forgot to check freshness" bugs). Step 2 would prevent R3b/R4a/R5a (the "filter-X-doesn't-do-Y" bugs).

**Wholesale rewrite (Redux/state machine)**: not justified. The OJS-imperative idiom is the project's convention; readers expect it.

### Bottom-line answer

A bit of all three, but mostly the third. Six rounds reflect Codex doing what humans rarely do — re-reading the diff after each fix and spotting the *next* missed edge in the same family. We were genuinely converging, just not in straight-line steps. Worth filing this small follow-up so future selection-state work can't trip the same wire.

## Refs

- PR #186 (cluster URL state, where the 6-round loop happened)
- `EXPLORER_STATE.md` (current state contract)
- `EXPLORER_CLUSTER_URL_PROPOSAL.md` (companion design doc)

## Next step

Codex's independent take will be posted as a comment. Then discussion + decision on whether to scope the refactor (and at what size) before any code.

rdhyee:
## Summary

Step 1 of #187 (post-mortem on PR #186's 6-round Codex loop). Behavior-preserving refactor.

## What changed

The same class of bug surfaced eight times across six review rounds: async work mutating selection state without checking that a newer user event had already superseded it. The duct-tape was a `_selGen` counter inlined at three call sites with slightly different shapes — each site bumped the counter, captured the snapshot, and rolled its own stale-check arrow.

This PR extracts that pattern as `freshSelectionToken()` (~10 lines) in the zoomWatcher cell:

```js
function freshSelectionToken() {
    viewer._selGen = (viewer._selGen || 0) + 1;
    const gen = viewer._selGen;
    return () => gen !== viewer._selGen;
}
```

Each user-input event handler that touches selection now does:

```js
const isStale = freshSelectionToken();
...
await someWork();
if (isStale()) return;
```

`isStale` is also passed through `hydrateClusterUI`'s existing predicate parameter so nested awaits bail before touching the DOM.

## Sites refactored

| Site | Before | After |
|---|---|---|
| Source-filter handler | `viewer._selGen = (viewer._selGen \|\| 0) + 1; const filterSelGen = viewer._selGen;` + 4 `filterSelGen !== viewer._selGen` checks | `const isStale = freshSelectionToken();` + 4 `isStale()` checks |
| Hashchange handler | inline bump + `const selGen = viewer._selGen;` + 2 manual checks | `const isStale = freshSelectionToken();` + 2 `isStale()` checks |
| Boot deep-link | inline bump + `const bootSelGen = ...; const isBootStale = () => ...;` + 4 `isBootStale()` checks | `const isStale = freshSelectionToken();` + 4 `isStale()` checks |

After the refactor, raw `_selGen` mentions exist only inside the helper. No other change in behavior.

## Doc update

`EXPLORER_STATE.md §4` adds:
- A row for `viewer._selGen` (was previously undocumented, slipped in over the 6 rounds).
- A new **Async-selection invariant** subsection that pins the rule: any async work that updates `_globeState` / URL hash / side-panel DOM must check `isStale()` after every await.

## YAGNI

Step 2 of #187 (consolidating selection mutations into a `selectSample` / `selectCluster` / `clearSelection` controller) is **deferred** per the synthesis with Codex on that issue — only worth doing if the next selection feature again has to touch click + boot + hashchange + filter + URL. Three call sites isn't enough to force consolidation. A wholesale state-machine rewrite is not justified.

## Verified locally

Re-ran the same three-step regression from PR #186's verification:

| step | result (vs. pre-refactor) |
|---|---|
| Boot at `#h3=843f6d3ffffffff` | identical: cluster card populated, 30 samples |
| Uncheck SESAR (cluster's source OpenContext stays) | identical: cluster card unchanged, samples re-rendered, `&h3=` preserved |
| Uncheck OpenContext (cluster's own source) | identical: card cleared, samples cleared, `&h3=` dropped |

## Out of scope

- Playwright stale-async regression test (deferred per #187 synthesis — timing-flaky in headless; manual DevTools recipe is enough until something forces our hand).
- Halo crescent / disappearing-dots over terrain (#185 — separately tracked).

Refs #187 (post-mortem + Codex synthesis), #186 (origin PR), `EXPLORER_STATE.md` (state contract).


rdhyee (commented):
Finding:

- P2: `freshSelectionToken()` does not cover the click path that the new invariant says it covers. `EXPLORER_STATE.md` says cluster/sample click handlers call `freshSelectionToken()`, but the helper is defined inside the later `zoomWatcher` cell, while the Cesium click handler is in the earlier `viewer` cell. The click handler still has async awaits with no freshness check: sample detail load and cluster nearby-samples load. A slow cluster click can still call `updateSamples(samples)` after the user has clicked a different sample/cluster. Either move the primitive to shared helper scope and use it in the viewer click handler, or narrow the docs/claim and leave click as explicit follow-up.

Verification: `quarto render explorer.qmd` and `git diff --check origin/main...HEAD` both passed.

rdhyee:
Pushed \`797605a\` addressing Codex's P2 finding:

The original PR put \`freshSelectionToken()\` in the zoomWatcher cell, but the cluster/sample click handler lives in the *earlier* viewer cell and couldn't see it — yet \`EXPLORER_STATE.md\`'s invariant claimed coverage of clicks. Two unguarded awaits in the click handler (sample detail load, cluster nearby-samples load) were the actual race surfaces: a slow earlier click could repaint the side panel after the user clicked a different sample/cluster.

Fix:
- Move \`freshSelectionToken(v)\` to top-level OJS scope alongside \`readHash\`/\`buildHash\`. Takes \`viewer\` as a parameter (defined before the viewer cell runs).
- Apply \`isStale()\` to the click handler's two await sites — sample-detail and cluster-nearby. Same shape as the other four sites.
- Updated \`EXPLORER_STATE.md\` invariant text to reflect the top-level definition.

Behavior-preserving regression check passes (PR #186's three-step suite returns identical results).

This makes the fifth call site of \`freshSelectionToken\` and the seventh \`isStale()\` check. The doc claim now matches the code.

rdhyee (commented):
Finding:

- P3: `EXPLORER_STATE.md` has one stale location note for `_selGen`. The row for `viewer._selGen` still says `freshSelectionToken()` is “in zoomWatcher cell,” but the latest patch correctly moved it top-level so both the viewer click handler and `zoomWatcher` can use it. The invariant text below is already updated; this row just needs the same wording.

No runtime issues found in the re-review. The click path now uses `freshSelectionToken(v)` and guards both async repaint sites.

Verification: `quarto render explorer.qmd` and `git diff --check origin/main...HEAD` both passed.

PR #195, opened 2026-05-10
explorer: document loadRes-chase invariant on tryEnterPointModeIfNeeded (closes #194)

### LINKED ISSUE #194: explorer: document the 'every loadRes caller chases with tryEnterPointModeIfNeeded' invariant

## Context

`tryEnterPointModeIfNeeded()` (introduced in #191) short-circuits when `currentRes !== 8 && loading === true`:

```js
async function tryEnterPointModeIfNeeded() {
    if (mode === 'point') return;
    if (viewer.camera.positionCartographic.height >= ENTER_POINT_ALT) return;

    let res8Ready = currentRes === 8;
    if (!res8Ready && !loading) {        // ← here
        res8Ready = await loadRes(8, h3_res8_url, { ... });
    }
    ...
}
```

This is safe in the current code because *every* `loadRes` call site in the cell is followed by an `await tryEnterPointModeIfNeeded()`, so any in-flight load that prevents this call from acting will itself trigger a chase when it settles.

But the safety relies on this invariant — and a future contributor adding a new `loadRes` call site without the chase would silently break supersession recovery. The bug would only surface as a rare liveness regression that's hard to attribute back to the missing chase.

## Proposed change

One-line comment near the `!loading` guard plus a short note in the helper's existing doc-block:

```js
async function tryEnterPointModeIfNeeded() {
    if (mode === 'point') return;
    if (viewer.camera.positionCartographic.height >= ENTER_POINT_ALT) return;

    let res8Ready = currentRes === 8;
    // Invariant: when an unrelated loadRes is in flight (`loading === true`),
    // we leave recovery to that call site's own post-await chase. Every
    // loadRes caller in this cell must call tryEnterPointModeIfNeeded()
    // afterwards — otherwise this short-circuit can strand the user.
    if (!res8Ready && !loading) {
        ...
    }
    ...
}
```

And a parallel sentence in the doc-block at the top of the helper.

## Verification

`grep -n 'loadRes(' explorer.qmd` should return only call sites that are immediately followed by `await tryEnterPointModeIfNeeded()`, OR are inside `tryEnterPointModeIfNeeded()` itself, OR are in code paths where the user can't be at point altitude (e.g. inside `exitPointMode` flow).

## Refs

- #190 — parent issue
- #191 — introduced `tryEnterPointModeIfNeeded()`
- [PR #191 final review comment](https://github.com/isamplesorg/isamplesorg.github.io/pull/191#issuecomment-4415443793) — flagged this invariant as load-bearing-but-undocumented

rdhyee:
## Summary

Closes #194. Doc-only change to `explorer.qmd` — adds a load-bearing **INVARIANT** block to the doc comment of `tryEnterPointModeIfNeeded()` plus a short pointer above the `!loading` short-circuit inside the function.

The invariant: every `loadRes` call site in the `zoomWatcher` cell that could be in flight when the user crosses below `ENTER_POINT_ALT` MUST be followed by `await tryEnterPointModeIfNeeded()`. Without the chase, the helper's `!loading` bail-out can strand the user in cluster mode at point altitude — the exact bug round 3 of #191 fixed by adding the source-filter handler's chase call.

## Why this matters

Future contributors adding a new `loadRes` site without the chase would silently break supersession recovery. The bug would surface only as a rare liveness regression (user at point altitude, mode stays cluster, no further events fire) that's hard to attribute back to the missing chase. The PR thread for #191 documented this; this PR moves it to the code itself.

## Test plan

- [ ] No behavior change — `quarto render explorer.qmd` produces identical output to main.
- [ ] `grep -nE 'loadRes\(' explorer.qmd` returns the same set of call sites as main, each either inside `tryEnterPointModeIfNeeded` itself or followed by `await tryEnterPointModeIfNeeded()`.

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1629,17 +1629,18 @@ zoomWatcher = {
     // or another path enters point mode during the wait, we must not force
     // entry afterwards.
     //
-    // INVARIANT: every `loadRes` call site in this cell that could be in
-    // flight when the user crosses below `ENTER_POINT_ALT` MUST be followed
-    // by `await tryEnterPointModeIfNeeded()`. The `!loading` short-circuit
-    // below relies on this — when an unrelated `loadRes` is in flight we
-    // bail and leave recovery to that call site's own post-await chase. A
-    // new `loadRes` caller added without the chase would silently break
-    // supersession recovery and only surface as a rare liveness regression
-    // (see issue #194). Verify with:
-    //   grep -nE "loadRes\(" explorer.qmd
-    // and check each match either lives inside `tryEnterPointModeIfNeeded`
-    // itself or is followed by a chase call.
+    // INVARIANT: every `loadRes` call site in this cell *outside this
+    // helper* that could be in flight when the user crosses below
+    // `ENTER_POINT_ALT` MUST be followed by `await tryEnterPointModeIfNeeded()`.
+    // The `!loading` short-circuit below relies on this — when an unrelated
+    // `loadRes` is in flight we bail and leave recovery to that call site's
+    // own post-await chase. A new `loadRes` caller added without the chase
+    // would silently break supersession recovery and only surface as a rare
+    // liveness regression (see issue #194). Verify with:
+    //   grep -nE "await[[:space:]]+loadRes\(" explorer.qmd
+    // which lists exactly the awaited call sites (one inside this helper,
+    // and one per external caller). For each external match, confirm it is
+    // immediately followed by `await tryEnterPointModeIfNeeded()`.
     async function tryEnterPointModeIfNeeded() {
         if (mode === 'point') return;
         if (viewer.camera.positionCartographic.height >= ENTER_POINT_ALT) return;
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1628,11 +1628,26 @@ zoomWatcher = {
     // long) `loadRes` await for the same reason: if the user zooms back out
     // or another path enters point mode during the wait, we must not force
     // entry afterwards.
+    //
+    // INVARIANT: every `loadRes` call site in this cell that could be in
+    // flight when the user crosses below `ENTER_POINT_ALT` MUST be followed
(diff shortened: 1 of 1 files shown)


rdhyee:
## Codex review round 1

Codex flagged one blocking issue + offered a precision tweak.

### Blocking

> The suggested `grep -nE "loadRes\(" explorer.qmd` matches comments at L1615 and L1623 (which mention `loadRes(...)` in prose), not just call sites. Misleading for someone running the verification.

**Fix in [b87cb49](https://github.com/rdhyee/isamplesorg.github.io/commit/b87cb49):** changed the documented command to `grep -nE "await[[:space:]]+loadRes\(" explorer.qmd`, which matches exactly four lines — the helper's own internal call plus the three external callers — with no false-positive comment matches.

Verified locally:
```
$ grep -nE "await[[:space:]]+loadRes\(" explorer.qmd
1653:            res8Ready = await loadRes(8, h3_res8_url, {       (inside the helper)
1830:                await loadRes(currentRes, resUrls[currentRes]);  (source-filter handler — chase ✓)
1944:                    await loadRes(target, ...)[target]);          (camera→cluster reload — chase ✓)
1957:                    await loadRes(target, ...)[target]);          (camera→cluster-resolution change — chase ✓)
```

### Precision tweak (also addressed)

> Optional: say "every `loadRes` call site outside this helper..." to avoid the helper's own internal call looking like an exception to the invariant sentence.

Added the "outside this helper" qualifier in the same commit.

### Codex's other answers (no action needed)

- ✓ Invariant matches the actual current call sites.
- ✓ Prose is clear without PR #191 context. "Chase" is local jargon but explained.

Asking for re-review.

rdhyee:
## Codex review round 2 — LGTM

> No blocking issues found.
>
> The round-2 fix resolves the round-1 finding: the updated grep now matches only awaited `loadRes(...)` call sites, not explanatory comment mentions. I verified it returns the expected four lines, and the three external call sites are immediately followed by `await tryEnterPointModeIfNeeded()`.
>
> No new issues introduced by round 2.
>
> **LGTM, ready to merge.**

Merging.

PR #196, opened 2026-05-10
explorer: skip tryEnterPointModeIfNeeded chase on non-applied loadRes (closes #193)

### LINKED ISSUE #193: explorer: smooth failure-message flicker when tryEnterPointModeIfNeeded chases a failed cluster reload

## Symptom

If a cluster reload fails (e.g. flaky network on the res4/res6 parquet) and the user is at point-mode altitude, the failed reload's call site immediately calls `tryEnterPointModeIfNeeded()`, which paints `Fetching sample index…` over the just-painted `Failed to load H3 res\${target} — try zooming again.` failure message. The user sees the failure flash for a frame, then the helper's loading-state takes over.

This is rare (it requires a cluster-reload failure on a path the camera handler is in, with the user already below `ENTER_POINT_ALT`) and self-recovering — but it's noisy when it does happen.

## Reproduction (synthetic)

1. Throttle network so res4/res6 parquet fetch fails.
2. Zoom from world view straight into point altitude (cluster reload fires, fails).
3. Observe phase-message strip during the transition.

## Possible fixes

A few shapes worth considering:

1. **Don't chase after a failed `loadRes`.** The chase is a *liveness* fix — recovering from supersession that left us silent. A failure already wrote a phase message, so silence isn't the issue. Skip the chase when `res8Ready === false`:

   ```js
   const ok = await loadRes(target, ...);
   if (ok) await tryEnterPointModeIfNeeded();
   ```

   Tradeoff: distinguishing supersession from failure at call sites where they're conflated (both return `false`). Could expand `loadRes`'s return to a tri-state (`'applied' | 'stale' | 'failed'`) but that's invasive.

2. **Brief debounce inside `tryEnterPointModeIfNeeded`.** If the helper is called within ~100ms of a phase-message write that wasn't a 'done' state, hold off the loading-message overwrite for that window so the failure remains visible long enough for the user to register it. Heuristic; brittle.

3. **Two-line phase-message slot.** Show last error (sticky for N seconds) on a second line instead of overwriting it. Bigger UI change; out of scope for a flicker fix.

Recommendation: **(1) with the tri-state return** is the cleanest, but only if other call sites also benefit. Otherwise, a one-line `if` at the chase sites checking `ok` is enough.

## Acceptance

- Failure messages from cluster-reload paths persist long enough to be human-readable (or at least: are not immediately overwritten by `Fetching sample index…` in the failure case).
- Supersession recovery (the actual round-3 fix) still works — toggle source filter during cold-cache wait still ends up in point mode.

## Refs

- #190 — parent issue
- #191 — introduced `tryEnterPointModeIfNeeded()`
- [PR #191 final review comment](https://github.com/isamplesorg/isamplesorg.github.io/pull/191#issuecomment-4415443793) — flagged this edge case

rdhyee:
## Summary

Closes #193. Stops `tryEnterPointModeIfNeeded()` from being invoked after a `loadRes` call returns `false`, so the user-facing failure phase message (`Failed to load H3 res${target} — try zooming again.`) no longer flickers before being overpainted by the chase's own `Fetching sample index…`.

## Root cause

After PR #191, three external `loadRes` call sites unconditionally chased with `await tryEnterPointModeIfNeeded()`:

```js
await loadRes(target, ...);
await tryEnterPointModeIfNeeded();   // unconditional
```

If `loadRes` *failed* (caught error → returned `false`), it painted `Failed to load H3 res${target} — try zooming again.` But the immediate chase, if the user is at point altitude, fired its own `loadRes(8, ...)` which painted `Fetching sample index…` over the failure. The user saw the failure message for ~1 frame.

If `loadRes` was *superseded* (stale → also `false`), the chase was technically harmless but redundant — the supersedor's own chase recovers the user.

## Fix

Capture the return into `applied` and gate the chase:

```js
const applied = await loadRes(target, ...);
if (applied) await tryEnterPointModeIfNeeded();
```

- **`applied === true`**: fresh data applied; chase fires as before.
- **`applied === false` (stale)**: supersedor's own chase will handle recovery (per the invariant documented in #194). Skipping our chase is a no-op for liveness.
- **`applied === false` (failed)**: failure phase message stays visible; user can read and act on it.

Applied at the three external call sites:
1. Source-filter handler's `mode === 'cluster'` branch
2. Camera handler's cluster-reload branch (`targetMode === 'cluster' && mode !== 'cluster'`)
3. Camera handler's cluster-mode resolution-change branch (`else` branch)

## Documentation

Updates the `tryEnterPointModeIfNeeded()` doc-block (added in #194 / #195) to:
- Show the new idiom: `const applied = await loadRes(...); if (applied) await tryEnterPointModeIfNeeded();`
- Explain *why* we gate on `applied` (one bullet per false-stale and false-failed case)
- Update the verification-grep instructions to point reviewers at the new shape

## Test plan

- [ ] **Cold-cache deep-link** still works (the original #190 fix 2 path). `Fetching sample index…` → `Loading individual samples...` → dots.
- [ ] **Source-filter toggle during cold-cache wait** still recovers into point mode (the round-3 fix in #191 — `applied=true` for the source-filter handler's loadRes(currentRes), so chase fires).
- [ ] **Synthetic res4/res6 fetch failure while crossing into point altitude**: throttle network, deep-zoom across cluster thresholds straight into point altitude. Confirm `Failed to load H3 res${target} — try zooming again.` stays visible (does not flicker to `Fetching sample index…`).
- [ ] **Unrelated regressions**: confirm existing cluster→cluster resolution changes still update the cluster overlays (they should — `applied=true` on success, chase only matters at point altitude).

## Refs

- #193 — this issue
- #194 / #195 — the loadRes-chase invariant (this PR updates the same doc-block to reflect the conditional chase idiom)
- #190 / #191 — original PR thread for the supersession recovery this builds on

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1631,16 +1631,32 @@ zoomWatcher = {
     //
     // INVARIANT: every `loadRes` call site in this cell *outside this
     // helper* that could be in flight when the user crosses below
-    // `ENTER_POINT_ALT` MUST be followed by `await tryEnterPointModeIfNeeded()`.
+    // `ENTER_POINT_ALT` MUST capture the call's return value and chase
+    // with `await tryEnterPointModeIfNeeded()` *iff that return is `true`*
+    // (i.e., fresh data was applied). Idiom:
+    //
+    //   const applied = await loadRes(...);
+    //   if (applied) await tryEnterPointModeIfNeeded();
+    //
     // The `!loading` short-circuit below relies on this — when an unrelated
     // `loadRes` is in flight we bail and leave recovery to that call site's
-    // own post-await chase. A new `loadRes` caller added without the chase
-    // would silently break supersession recovery and only surface as a rare
-    // liveness regression (see issue #194). Verify with:
+    // own post-await chase.
+    //
+    // Why gate the chase on `applied` (issue #193):
+    //  - On `false`-stale: a newer `loadRes` caller superseded us; that
+    //    caller chases when it settles, so our chase would be redundant.
+    //  - On `false`-failed: `loadRes` already painted a "Failed to load…"
+    //    phase message that the user should be able to read; chasing here
+    //    would immediately overpaint it with "Fetching sample index…".
+    //
+    // A new `loadRes` caller added without the chase would silently break
+    // supersession recovery and only surface as a rare liveness regression
+    // (see issue #194). Verify with:
     //   grep -nE "await[[:space:]]+loadRes\(" explorer.qmd
     // which lists exactly the awaited call sites (one inside this helper,
-    // and one per external caller). For each external match, confirm it is
-    // immediately followed by `await tryEnterPointModeIfNeeded()`.
+    // and one per external caller). For each external match, confirm it
+    // captures the return into `applied` and is immediately followed by
+    // `if (applied) await tryEnterPointModeIfNeeded();`.
     async function tryEnterPointModeIfNeeded() {
         if (mode === 'point') return;
         if (viewer.camera.positionCartographic.height >= ENTER_POINT_ALT) return;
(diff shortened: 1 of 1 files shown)


rdhyee:
## Codex review round 1 — LGTM

> **Findings:** None. I don't see a blocking counterexample.
>
> Verified against `HEAD` on `fix/193-skip-chase-on-loadres-failure`:
>
> 1. The stale-recovery claim is sound for the current code. A stale `loadRes` return means a newer `loadRes` incremented `loadResGen`; the only live external superseding sites are the three edited sites, and each now chases when its own load applies. The helper's internal res8 load will not start while another `loadRes` is in flight because of the `!loading` guard.
> 2. Yes, this fixes the failure-message flicker path. `loadRes` paints the failure message and returns `false`; the three external callers now skip `tryEnterPointModeIfNeeded()` on that `false`, so they no longer immediately trigger the helper's "Fetching sample index…" path.
> 3. The edits are complete. `rg "await\s+loadRes\(" explorer.qmd` shows one internal helper call and three external call sites — all three external sites have the new `const applied = ...; if (applied) await tryEnterPointModeIfNeeded();` shape.
> 4. The doc block is consistent with the code.
> 5. No new issue introduced. The only non-recovery case left is an actual newest-load failure, which is intentional so the user can read the failure message.
>
> `git diff --check upstream/main..HEAD` passed.
>
> **LGTM, ready to merge.**

Merging.

PR #202, opened 2026-05-11
[codex] Stabilize explorer map layout

### LINKED ISSUE #199: explorer: map position and size jump between views because surrounding elements shift
## Symptom

The Cesium globe (`#cesiumContainer`) changes both **vertical position on the page** and **rendered size** as the user moves between views (cold-load, search-active, point-mode, cluster-mode, table-mode). The user has to re-orient on the map each time something above or beside it changes.

## Diagnosis

Two distinct causes, both in `explorer.qmd`:

### 1. Map size is content-width-dependent

`#cesiumContainer` (`explorer.qmd:40-44`) is currently sized as:

```css
#cesiumContainer {
    width: 100%;
    min-height: 500px;
    aspect-ratio: 4/3;
}
```

Height is derived from width via aspect-ratio. Width is the left column of `grid-template-columns: 1fr 340px` on `.globe-layout`. So any viewport-width change resizes the map.

The side-panel column has `max-height: 700px; overflow-y: auto`, which is the right pattern (bounded + internal scroll), but the map column has no comparable height anchor.

### 2. Map *position* drifts because variable-height content lives above it

The flow above `.globe-layout`:

- `.search-bar` — fixed
- `.search-actions` — fixed
- `.search-help` — fixed
- `#searchResults` (`.search-results`) — **variable**, grows with results
- `.view-toolbar` — fixed, but `#tableControls` toggles visibility
- Phase / "Fetching sample index…" / error banners (added in #190 / PR #191) — appear and disappear

When any of those grow/shrink, `#cesiumContainer.getBoundingClientRect().top` shifts and the map jumps vertically.

## Proposed fix (CSS + light HTML restructure)

Treat the globe as an **app-shell anchor**, not a flow-layout element:

1. Replace `aspect-ratio: 4/3` + `min-height: 500px` on `#cesiumContainer` with a viewport-stable height (e.g., `height: clamp(500px, 70vh, 800px)`). Map height becomes a function of viewport height alone.
2. Bound `#searchResults` with `min-height: 0; max-height: <small>; overflow-y: auto` — or move inline status into a Cesium overlay (`.cesium-topleft` slot already styled at `explorer.qmd:23-30`).
3. Reserve a **fixed-height slot** for banner-style messages above the map (loading, phase, error) so the slot exists whether or not a message is showing.
4. Verify with Playwright: snapshot `#cesiumContainer.getBoundingClientRect()` across the five views; assert top/height stable within ε.

No JS logic changes expected. Scope is similar to a single Codex-loop PR.

## Open design choices

- **Target height**: `clamp(500px, 70vh, 800px)` is a starting point. May want to align with the Hana mockup direction referenced near `.cesium-viewer-toolbar` styling (`explorer.qmd:~98`).
- **Status banners — overlay vs reserved slot**: overlay (Cesium top-left badge, reusing `.cesium-topleft` styling) is denser; reserved slot is more conventional. PR #191's "Fetching sample index…" UX is currently implemented one of these ways — choice should be consistent with that.

## Acceptance criteria

- [ ] `#cesiumContainer.getBoundingClientRect()` has constant top-y and height across cold-load, search-active, point-mode, cluster-mode, and table-mode (within a small ε).
- [ ] Window resize changes map size monotonically with viewport height, not with side-panel content state.
- [ ] No regression in the cold-cache UX from #190 / PR #191 (loading message still visible, doesn't itself shift the map).
- [ ] Playwright spec covering the stable-geometry invariant.

rdhyee:
## Summary

Fixes #199.

This stabilizes the explorer globe layout so the Cesium container is anchored to viewport height instead of deriving its height from grid-column width. It also reserves bounded space for search status text and the view toolbar so loading/search/table controls cannot move the map frame.

## Changes

- Replaced `#cesiumContainer` aspect-ratio sizing with `--explorer-map-height: clamp(500px, 65vh, 680px)`.
- Added a mobile map-height override of `clamp(360px, 58vh, 520px)`.
- Matched `.side-panel` max height to the same map-height variable.
- Added `scrollbar-gutter: stable` and `minmax(0, 1fr)` for layout-width stability.
- Wrapped above-map controls in `.explorer-controls`, reserved a two-line `#searchResults` live region, and reserved view-toolbar height so table controls do not shift the table/globe anchor.
- Added Playwright coverage for desktop and mobile layout stability.

## Validation

- `node --check tests/playwright/explorer-layout-stability.spec.js`
- `env NODE_PATH=/Users/raymondyee/C/src/iSamples/isamplesorg.github.io/node_modules /Users/raymondyee/C/src/iSamples/isamplesorg.github.io/node_modules/.bin/playwright test tests/playwright/explorer-layout-stability.spec.js --project=chromium --reporter=line`

The focused Playwright spec passed: 2 tests passed on Chromium.

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -17,6 +17,7 @@ format:
         <link rel="preload" as="fetch" crossorigin="anonymous" href="https://data.isamples.org/vocab_labels.parquet">
 ---
 
+```{=html}
 <script src="https://cesium.com/downloads/cesiumjs/releases/1.127/Build/Cesium/Cesium.js"></script>
 <link href="https://cesium.com/downloads/cesiumjs/releases/1.127/Build/Cesium/Widgets/widgets.css" rel="stylesheet"></link>
 <style>
@@ -279,6 +280,7 @@ format:
     .facet-row.zero:hover { opacity: 0.65; }
     .facet-count.recomputing { opacity: 0.55; font-style: italic; }
 </style>
+```
 
 ::: {.callout-note collapse="true"}
 ## How It Works
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -23,9 +23,16 @@ format:
 <style>
     :root {
         --explorer-map-height: clamp(500px, 65vh, 680px);
+        --explorer-shell-width: min(1420px, calc(100vw - 64px));
     }
     html {
         scrollbar-gutter: stable;
+    }
+    #quarto-document-content {
+        max-width: none;
+        width: var(--explorer-shell-width);
+        margin-left: 50%;
+        transform: translateX(-50%);
     }
     div.cesium-topleft {
         display: block;
@@ -44,6 +51,7 @@ format:
     @media (max-width: 900px) {
         :root {
             --explorer-map-height: clamp(360px, 58vh, 520px);
+            --explorer-shell-width: calc(100vw - 32px);
         }
         .globe-layout { grid-template-columns: 1fr; }
     }
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -20,6 +20,12 @@ format:
 <script src="https://cesium.com/downloads/cesiumjs/releases/1.127/Build/Cesium/Cesium.js"></script>
 <link href="https://cesium.com/downloads/cesiumjs/releases/1.127/Build/Cesium/Widgets/widgets.css" rel="stylesheet"></link>
 <style>
+    :root {
+        --explorer-map-height: clamp(500px, 65vh, 680px);
+    }
+    html {
+        scrollbar-gutter: stable;
+    }
     div.cesium-topleft {
         display: block;
         position: absolute;
@@ -30,23 +36,27 @@ format:
     }
     .globe-layout {
         display: grid;
-        grid-template-columns: 1fr 340px;
+        grid-template-columns: minmax(0, 1fr) 340px;
         gap: 12px;
         margin-bottom: 16px;
     }
     @media (max-width: 900px) {
+        :root {
+            --explorer-map-height: clamp(360px, 58vh, 520px);
+        }
         .globe-layout { grid-template-columns: 1fr; }
     }
     #cesiumContainer {
         width: 100%;
-        min-height: 500px;
-        aspect-ratio: 4/3;
+        height: var(--explorer-map-height);
+        min-height: 0;
+        aspect-ratio: auto;
     }
     .side-panel {
         display: flex;
         flex-direction: column;
         gap: 8px;
-        max-height: 700px;
+        max-height: var(--explorer-map-height);
         overflow-y: auto;
     }
     .panel-section {
@@ -111,14 +121,20 @@ format:
         left: 36px;
         right: auto;
     }
-    .search-bar { display: flex; gap: 6px; margin-bottom: 6px; }
+    .explorer-controls {
+        display: flex;
+        flex-direction: column;
+        gap: 6px;
+        margin-bottom: 12px;
+    }
+    .search-bar { display: flex; gap: 6px; }
     .search-bar input {
         flex: 1; padding: 8px 12px; border: 1px solid #ccc; border-radius: 4px;
         font-size: 14px; outline: none;
     }
     .search-bar input:focus { border-color: #1565c0; box-shadow: 0 0 0 2px rgba(21,101,192,0.15); }
     .search-actions {
-        display: flex; gap: 6px; margin-bottom: 8px;
+        display: flex; gap: 6px;
     }
     .search-actions button {
         flex: 1; border: none; padding: 8px 12px; border-radius: 4px;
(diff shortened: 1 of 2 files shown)


rdhyee:
## Codex review — round 1

Self-contained Codex round-1 review (no shared context with author). Verbatim per the iterative-review pattern.

---

## Summary
The fixed map-height strategy is sound, and it directly addresses the vertical jump described in #199. I would not merge this exact revision yet: the rendered Quarto HTML does not match the control DOM the CSS assumes, and the new Playwright spec has the known brittle point-mode settle wait.

## Per-ask response

### 1. Quarto raw-html fence
The fence is right for the `<script>/<link>/<style>` block. Bare HTML is preserved, but Quarto still parses Markdown inside it: rendered `docs/explorer.html` wraps several controls in `<p>` tags, e.g. `.search-bar > p > input` and `#tableControls > p > label + input`.

That matters because `.search-bar input { flex: 1 }` no longer makes the input a flex item, and paragraph margins feed into toolbar/control heights. I would also fence the static app markup, or explicitly neutralize those wrappers.

### 2. `#quarto-document-content` override
The selector is probably stable enough for Quarto, but it is broad: title, callout, code blocks, and the app all inherit the wider shell. `margin: 0 auto` is probably not equivalent inside Quarto’s page grid; the transform is the escape hatch. Cleaner options are Quarto `page-layout: full` or a page-local app wrapper.

### 3. Reserved search slot
Two lines covers normal messages: “Searching…”, “No results…”, and short result counts. Long user queries and `Search error: ${err.message}` can exceed it and will scroll. That is acceptable for layout stability, but three lines would be more forgiving.

### 4. `.view-toolbar { min-height: 50px }`
Likely enough on desktop. Risky on narrow mobile when `#tableControls` is visible and wraps, especially with the rendered `<p>` wrapper. The mobile spec does not exercise table mode, so this gap is currently untested.

### 5. Map height vs side panel
No functional issue. CSS Grid’s default stretch should make the side panel occupy the row height up to the matching `max-height`; if content is short, it may look visually sparse because the panel itself has no background.

### 6. Mobile coverage
The 390×844 math is correct. I would add a small-phone case, e.g. 320×568 or 390×600, mainly to cover the 360px floor and control wrapping behavior.

### 7. Spec sample cap brittleness
Yes, switch the wait from `'individual samples'` to `'Click one for details'`. The cap branch in `explorer.qmd` omits “individual”, so this can time out under dense views or future data changes.

### 8. `waitForClusterBoot` / `zoomWatcher`
Not a no-op. Awaiting `zoomWatcher` ensures the OJS cell has completed listener registration and boot hydration. A comment would help; alternatively wait for `_suppressHashWrite === false`.

### 9. `requestRenderMode = false`
Fine for Playwright. Each test gets a fresh page, and this matches existing URL-roundtrip practice.

### 10. `aria-live="polite"`
No major concern. Search status changes on explicit search actions, not every keystroke, so it should not be spammy.

### 11. Other
The largest missing test assertion is the rendered control layout: search input width and mobile table-mode top stability. Also consider deriving expected map heights from computed CSS instead of hard-coded `585` / `489.52`.

## Verdict
BLOCK

rdhyee:
## Codex review — round 2

Verbatim Codex round-2 review. **LGTM, ready to merge.**

Codex's only non-blocking nit (consistency: desktop test still hardcoded 585 while mobile cases use the probe) addressed in a follow-up commit — extracted a `resolveMapHeightPx(page)` helper and used it from all three tests.

---

## Summary
No block-level issues found in the round-2 diff. The raw HTML fence boundaries look right, the probe-element approach is appropriate for resolving the CSS `clamp()`, and the small-phone test covers the intended floor branch.

I did not rerun Playwright locally; this is a static re-review of the checked-out files.

## Per-ask response

### 1. P-wrap fix scope
The boundary is right. Static interactive HTML is inside the two `{=html}` fences:

- [explorer.qmd](/private/tmp/isamplesorg-issue199/explorer.qmd:310): search controls, view toggle, `label`, and `maxSamples` input.
- [explorer.qmd](/private/tmp/isamplesorg-issue199/explorer.qmd:338): globe layout, side panel filters, share button, and table pager buttons.

The only other interactive-looking markup found by `rg` is inside OJS/JS template strings, so Pandoc will not p-wrap it.

### 2. Probe-element approach
Yes, this is the right idiom here. `getPropertyValue('--explorer-map-height')` returns the unresolved custom property, while a hidden element gives the used pixel height after `var()` + `clamp()` resolution.

The probe and `#cesiumContainer` should resolve identically for the current CSS because the value only depends on root variables and viewport units. They could diverge only if this later changes to context-dependent sizing, container units, inherited overrides, padding/borders, or min/max constraints on the real element.

### 3. Small-phone test correctness
Correct. At `320×568`, the mobile media query applies, `58vh = 329.44px`, and `clamp(360px, 58vh, 520px)` resolves to `360px`.

`expect(expectedMapHeight).toBe(360)` is fine in Chromium for this floor branch. A tolerance would be slightly more consistent with the rest of the spec, but I would not block on it.

### 4. PR title `[codex]`
I would strip `[codex]` before squash merge unless the repo intentionally uses that convention. Provenance is better kept in PR metadata; commit history usually reads better without tool prefixes.

### 5. Anything else
One non-blocking nit: the desktop assertion still hardcodes `585` at [explorer-layout-stability.spec.js](/private/tmp/isamplesorg-issue199/tests/playwright/explorer-layout-stability.spec.js:91). The mobile cases now use the probe, so the desktop case could use the same helper for consistency.

## Verdict
LGTM, ready to merge

PR #203, opened 2026-05-11
explorer: fix URL state round-trip (closes #201 parts A+B)

### LINKED ISSUE #201: explorer: 'Samples in View' shows the 5000 fetch budget, not real count + URL round-trip doesn't reproduce view
Investigated 2026-05-11 starting from this URL:
https://isamples.org/explorer.html#v=1&lat=34.9957&lng=33.6798&alt=15212&heading=360.0&mode=point

User-visible symptoms:
1. The "Samples in View" stat box reads exactly **5000** — a suspiciously round number.
2. Pan and zoom from that starting state, copy the URL, paste into a different browser → the view that comes back is not the view that was captured.

Both symptoms have concrete root causes. Treating as a single issue because they were investigated together and overlap; happy to split if a maintainer prefers.

---

### Root cause

`explorer.qmd:418`

```js
DEFAULT_POINT_BUDGET = 5000
```

```sql
  ${sourceFilterSQL('source')}
  ${facetFilterSQL()}
LIMIT 5000
```

`explorer.qmd:1557` — the UI counter:

```js
```

`cachedData` is the result of the `LIMIT 5000` query. The counter therefore tops out at 5000 by construction. In dense regions it does not represent "samples in view" — it represents "samples we chose to load from a slightly-padded box around the view."

### Ground-truth numbers for the Cyprus URL

Direct DuckDB query against `https://data.isamples.org/isamples_202601_samples_map_lite.parquet`, centered on the URL's `lat=34.9957, lng=33.6798`:

| Viewport (degrees half-extent) | Actual samples |
|---|---:|
| ±0.10° | **23,421** |
| ±0.20° | 23,803 |
| ±0.50° | 24,305 |
| ±1.00° | 37,869 |

So when the UI says "5,000 Samples in View" at Cyprus, the truth is 23,000+ even within a viewport tighter than the explorer's 30%-padded fetch box. Counter is wrong by ~5x in this region.

### Secondary smells in the same query path

- **No `ORDER BY` before `LIMIT 5000`** → which 5000 rows are returned is undefined. DuckDB-on-parquet is *probably* stable file-order in practice, but it's not a contract.
- **Label says "in View" but fetch uses a padded box** (30% larger; `explorer.qmd:1514-1522`). Even if we set aside the cap, the count is loosely defined.
- **The displayed count never shrinks to "true visible"** as the user pans inside the cached padded box — `renderSamplePoints` plots all of `cachedData`, including rows outside the actual viewport.

### What's at Cyprus (for context)

The 23K samples in the box are one dense cluster around lat 34.98, lng 33.71, all OPENCONTEXT. That's the Polis excavations project (Excavations at Polis had 52,762 OC records per the Open Context facet API). So the cap is hiding a single very-dense site, not a diffuse distribution.

### Suggested fix directions (for discussion, not prescriptive)

1. **Cheapest**: change the label from "Samples in View" to "Samples Loaded (max N)" so the counter no longer lies. Wire the budget value into the label.
2. **Show the real count separately**: a fast `SELECT count(*)` against the same WHERE clause (no LIMIT) is cheap on the lite parquet via DuckDB-WASM range reads. Display two numbers: real count and rendered count.
3. **Adaptive budget / aggregation**: if real count > budget, fall back to a server-side aggregation or show "23,421 samples — too dense to render individually" with a UI affordance to drill in.
4. **Add `ORDER BY pid` (or similar)** so the LIMIT 5000 subset is at least deterministic across browsers and sessions.

---

### What we know about the write path

Camera-change handler (`explorer.qmd:1965-2030`):

```js
viewer.camera.changed.addEventListener(() => {
    if (timer) clearTimeout(timer);
    timer = setTimeout(async () => {
        // ... mode/resolution decisions ...
        if (!viewer._suppressHashWrite) {
            history.replaceState(null, '', buildHash(viewer));
        }
    }, 600);
});
```

`buildHash` (`explorer.qmd:651-671`) encodes only: `v=1`, `lat`, `lng`, `alt`, optional `heading` (only if `abs(heading % 360) > 1`), optional `pitch` (only if not nadir), optional `mode=point`, optional `pid` or `h3`. The query-string state (`?search=`, `?sources=`, `?material=`, etc.) is written by a separate `writeQueryState()` function (`explorer.qmd:494-526`) on filter changes, not by the camera handler.

### Hypotheses (need empirical confirmation per the EXPLORER_STATE.md state-contract framing, #164)

1. **Copy mid-debounce.** 600ms debounce means a user who pans/zooms and immediately copies the URL gets stale state. Easy to confirm: pan, wait 2s, then check the address bar.
2. **Heading normalization drops `360.0`.** `buildHash` only writes heading if `abs(heading % 360) > 1` (line 661). The deep-link URL Raymond used has `heading=360.0`, but after one camera-handler tick that value is normalized to `0` and the param is dropped. If the user's view depends on heading != 0, the next URL write silently discards it.
3. **Cold-cache point-mode latency dominates.** Deep-link to `mode=point` triggers the res8 + samples_map_lite fetch path that takes 60–90s on a cold cache (the same path #190 / PR #191 worked around). In a fresh browser the view "looks different" might mean "samples haven't loaded yet."
4. **5000-cap non-determinism** (Part 1). Even when both browsers finish loading, the displayed 5000-sample subset is undefined without `ORDER BY`. The two browsers might render *different* 5000 dots.
5. **`_suppressHashWrite` could stay stuck.** Hashchange handler sets it true (line 2127) and clears it after a 2000ms timeout (lines 2140-2145). If a user chains hashchanges (back/forward repeatedly) faster than 2000ms, the flag may stay set across writes. Edge case; less likely in Raymond's flow but worth ruling out.

### Cleanest discriminating test

After pan/zoom and a 2-second pause:
- **If the address bar contains the current camera state** → problem is on the *load* side. Suspects: (3) cold-cache latency, (4) 5000-cap subset roulette, (5) stuck suppress flag.
- **If the address bar still has stale state** → problem is on the *write* side. Suspects: (1) longer-than-600ms thing keeping `_suppressHashWrite` true, or (2) heading-normalization dropping a param the user needed.

### Relationship to other work

- **#164** Explorer state contract: this issue is exactly the kind of empirical fodder #164's state inventory needs. Add a row for each URL-encoded field including "what triggers it" and "what swallows it" (e.g., heading-normalization-on-write is a *swallow* the inventory should capture).
- **#190 / PR #191**: confirms the cold-cache cost on point-mode deep-link is a known multi-second delay; hypothesis (3) is plausible.
- **#199**: layout-stability work that's in-flight separately.

---

## Repro (Cyprus, 2026-05-11)

```
URL: https://isamples.org/explorer.html#v=1&lat=34.9957&lng=33.6798&alt=15212&heading=360.0&mode=point
Observed: "Samples in View: 5,000"
Real count in tight ±0.1° box: 23,421 (one dense OPENCONTEXT cluster, probably Polis)
```

DuckDB reproduction (no auth needed):

```python
import duckdb
con = duckdb.connect()
con.execute("INSTALL httpfs; LOAD httpfs;")
url = 'https://data.isamples.org/isamples_202601_samples_map_lite.parquet'
con.execute(f"""
    SELECT count(*) FROM read_parquet('{url}')
    WHERE latitude BETWEEN 34.8957 AND 35.0957
      AND longitude BETWEEN 33.5798 AND 33.7798
""").fetchone()
# (23421,)
```

rdhyee:
Closes #201 (parts A + B). Part 1 — the misleading 'Samples in View' counter — is left for a separate PR.

## What this changes

Two small, independent fixes in `explorer.qmd`:

### Bug A — camera-handler URL write was deferred behind awaits

`explorer.qmd:1965-2030` — the camera-changed handler's debounced callback awaits `loadRes` / `tryEnterPointModeIfNeeded` (cold-cache point-mode boot can be 60–90s per #190) before reaching the trailing `history.replaceState`. When a user pans into a region that triggers a resolution change or point-mode transition, the URL write is deferred until those awaits settle. A URL copy in the meantime captures the **previous** camera state.

**Fix**: write the URL hash at the START of the debounced callback, before any awaits. The camera state is known synchronously. The trailing write is kept so any mode change produced by the awaits still gets captured.

### Bug B — deep-link `mode=point` silently ignored when heading=0

`explorer.qmd` boot path positions the camera via `camera.setView`, which in some cases (notably when the URL omits `heading=` and it defaults to 0) does not raise `camera.changed`. The camera-handler that drives the cluster→point transition therefore never runs, and the URL's `mode=point` is ignored.

**Fix**: explicit `tryEnterPointModeIfNeeded()` call at the end of the deep-link boot path when `ih.mode === 'point'`. The helper short-circuits if alt >= ENTER_POINT_ALT or we're already in point mode, so this is a no-op for cluster deep-links.

## Verification

`tests/playwright/url_roundtrip_investigation.js` is the harness that originally detected the bugs (see #201 comment with empirical evidence). It drives Context A through 6 deterministic pan/zoom steps from the Cyprus deep-link, then opens each resulting URL in a fresh Context B and probes at 5s/15s/30s.

| Iter | Pan/zoom step | Live site (pre-fix) | localhost (post-fix) |
|---|---|---|---|
| 1 | no-op baseline | ✅ | ✅ |
| 2 | pan NE only | ❌ URL stale (Bug A) | ✅ |
| 3 | zoom in, heading 0 | ❌ B stuck cluster (Bug B) | ✅ |
| 4 | heading 45° at zoom | ✅ | ✅ |
| 5 | pan+zoom+heading 90° | ✅ | ✅ |
| 6 | heading 360° (=0 mod) | ❌ B stuck cluster (Bug B) | ✅ |

Harness logs land at `/tmp/url_roundtrip_log.json`. Run with `TEST_URL=http://localhost:5860/explorer.html node tests/playwright/url_roundtrip_investigation.js`.

## Risks

- **Double URL write per camera-change**: The handler now writes the hash at the start AND at the end of the debounced callback. Both are `replaceState` (no history-entry growth). If the awaits don't change mode, the second write is identical to the first. Acceptable cost.
- **Bug B fix relies on `tryEnterPointModeIfNeeded()` being safe to call from boot.** Reviewed: it short-circuits on `mode === 'point'` and on alt >= ENTER_POINT_ALT, and uses the same `loadRes` path that the camera handler would. Functionally equivalent to "what would happen if `camera.changed` fired."
- **Cesium `setView`-doesn't-fire-camera.changed** is the underlying cause of Bug B's mechanism. Documented inline at the boot call site. This PR works around it; a longer-term fix would be to ensure `setView` raises the event (likely by tweaking the call shape or following with a no-op camera move).
- **No new tests in the regression suite yet.** The harness is shipped as an artifact under `tests/playwright/` but is a Node script, not a Playwright spec. A follow-up should adapt it into a proper `*.spec.js` so it runs in CI. Tracked as a TODO in the harness file's header comment.

## Test plan

- [x] Harness passes all 6 iterations against localhost (Quarto preview, this branch)
- [ ] Manual smoke on deploy preview: Cyprus deep-link → pan → copy URL → paste in private window → view matches
- [ ] Codex review (next round)

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1967,6 +1967,18 @@ zoomWatcher = {
         timer = setTimeout(async () => {
             const h = viewer.camera.positionCartographic.height;
 
+            // Write the URL hash BEFORE any awaits (issue #201, Bug A).
+            // The camera position/heading is known synchronously; without this
+            // early write, a user who pans into a region that triggers a
+            // resolution change (cluster→res8) or a cold-cache point-mode
+            // transition won't see the URL update until the await chain
+            // below settles — and a URL copy in the meantime captures the
+            // PREVIOUS state. We still write again at the end of the
+            // handler to capture any mode change that the awaits produced.
+            if (!viewer._suppressHashWrite) {
+                history.replaceState(null, '', buildHash(viewer));
+            }
+
             // Determine target mode with hysteresis
             const targetMode = h < ENTER_POINT_ALT ? 'point'
                              : h > EXIT_POINT_ALT ? 'cluster'
@@ -2582,6 +2594,19 @@ zoomWatcher = {
         // Enable hash writing now that everything is initialized — runs even on
         // stale-abort early returns above.
         viewer._suppressHashWrite = false;
+    }
+
+    // Deep-link mode hydration (issue #201, Bug B). The boot path positions
+    // the camera via `camera.setView`, which in some cases (notably when the
+    // URL omits `heading` — i.e. heading defaults to 0 — at point altitudes)
+    // does not raise `camera.changed`, so the camera-changed handler never
+    // runs and the URL's `mode=point` is silently ignored. Drive the mode
+    // transition explicitly here so the boot is independent of whether
+    // `setView` happened to fire the event. `tryEnterPointModeIfNeeded()`
+    // short-circuits if alt >= ENTER_POINT_ALT or we're already in point
+    // mode, so this is a no-op for cluster deep-links.
+    if (ih.mode === 'point') {
+        await tryEnterPointModeIfNeeded();
     }
 
     return "active";
--- a/tests/playwright/url_roundtrip_investigation.js
+++ b/tests/playwright/url_roundtrip_investigation.js
@@ -1,0 +1,224 @@
+/**
+ * URL state round-trip investigation (issue #201, Part 2).
+ *
+ * Hypothesis space (from #201):
+ *   1. Copy mid-debounce → URL stale
+ *   2. heading=360 normalized to 0 → param dropped
+ *   3. Cold-cache point-mode latency → "different view" = "samples not loaded yet"
+ *   4. 5000-cap with no ORDER BY → different 5000 across browsers
+ *   5. _suppressHashWrite stuck via rapid hashchanges
+ *
+ * Method:
+ *   - Context A loads the Cyprus deep-link, settles into point mode,
+ *     then performs N deterministic flyTo() pan/zoom steps. After each
+ *     settle, snapshots the URL + viewer state + cachedData length.
+ *   - For each snapshot, Context B opens a fresh page at that URL,
+ *     waits stable, snapshots the same fields. Compares.
+ *
+ * Output:
+ *   - /tmp/url_roundtrip_log.json — full structured log
+ *   - stdout summary of diffs
+ */
+
+const { chromium } = require('playwright');
+
+const SITE = process.env.TEST_URL || 'https://isamples.org/explorer.html';
+const START_HASH = '#v=1&lat=34.9957&lng=33.6798&alt=15212&heading=360.0&mode=point';
+const ITERATIONS = 6;
+const SETTLE_MS = 2500;          // after a camera move, wait for debounce + URL write
+const B_WAITS = [5000, 15000, 30000];  // probe Context B at multiple times
+
+/**
+ * Capture all the state we care about from a page that has booted the explorer.
+ * Returns null if the explorer isn't ready yet.
+ */
+async function capture(page, label) {
+  return await page.evaluate(async (label) => {
+    const v = await window._ojs?.ojsConnector?.mainModule?.value('viewer').catch(() => null);
+    if (!v || !v.camera || !v._globeState) return { label, ready: false, url: location.href };
+    const carto = v.camera.positionCartographic;
+    const cached = await window._ojs.ojsConnector.mainModule.value('cachedData').catch(() => null);
+    return {
+      label,
+      ready: true,
+      ts: Date.now(),
+      url: location.href,
+      hash: location.hash,
+      camera: {
+        lat: Cesium.Math.toDegrees(carto.latitude),
+        lng: Cesium.Math.toDegrees(carto.longitude),
+        alt: carto.height,
+        heading: Cesium.Math.toDegrees(v.camera.heading),
+        pitch: Cesium.Math.toDegrees(v.camera.pitch),
+      },
+      mode: v._globeState.mode,
+      selectedPid: v._globeState.selectedPid || null,
+      selectedH3: v._globeState.selectedH3 || null,
+      cachedDataLen: Array.isArray(cached) ? cached.length : null,
+      suppressHashWrite: !!v._suppressHashWrite,
+      selGen: v._selGen || null,
+      phaseMsg: document.getElementById('phaseMsg')?.textContent?.trim() || '',
+      sSamples: document.getElementById('sSamples')?.textContent?.trim() || '',
+    };
+  }, label);
+}
+
+/**
+ * Wait until viewer is initialized and we're in point mode (deep-link end state).
+ */
+async function waitReadyAndPointMode(page, timeoutMs = 120000) {
+  return page.waitForFunction(
+    async () => {
+      try {
+        const v = await window._ojs?.ojsConnector?.mainModule?.value('viewer');
+        if (!v?._globeState) return false;
+        return v._globeState.mode === 'point';
+      } catch { return false; }
+    },
+    null,
+    { timeout: timeoutMs }
+  );
+}
+
+/**
+ * Drive the camera with a deterministic flyTo, return after a settle period.
+ */
+async function flyAndSettle(page, target, settleMs = SETTLE_MS) {
+  await page.evaluate(async ({ lat, lng, alt, heading, pitch, durationS }) => {
+    const v = await window._ojs.ojsConnector.mainModule.value('viewer');
+    v.camera.cancelFlight();
+    v.scene.requestRenderMode = false;
+    v.camera.flyTo({
+      destination: Cesium.Cartesian3.fromDegrees(lng, lat, alt),
+      orientation: {
+        heading: Cesium.Math.toRadians(heading),
+        pitch: Cesium.Math.toRadians(pitch),
+      },
+      duration: durationS,
+    });
+  }, target);
+  // Wait flight duration + debounce + a margin
+  await page.waitForTimeout((target.durationS * 1000) + settleMs);
+}
+
+/** Diff two captures, return human-readable list of significant deltas. */
+function diff(a, b) {
+  const out = [];
+  if (!a?.ready || !b?.ready) {
+    out.push(`ready: A=${a?.ready} B=${b?.ready}`);
+    return out;
+  }
+  const dLat = Math.abs(a.camera.lat - b.camera.lat);
+  const dLng = Math.abs(a.camera.lng - b.camera.lng);
+  const dAlt = Math.abs(a.camera.alt - b.camera.alt);
+  const dHeading = Math.min(
+    Math.abs(a.camera.heading - b.camera.heading),
+    360 - Math.abs(a.camera.heading - b.camera.heading)
+  );
+  const dPitch = Math.abs(a.camera.pitch - b.camera.pitch);
+  if (dLat > 0.001) out.push(`lat Δ${dLat.toFixed(5)}`);
+  if (dLng > 0.001) out.push(`lng Δ${dLng.toFixed(5)}`);
+  if (dAlt > 50) out.push(`alt Δ${Math.round(dAlt)}m (A=${Math.round(a.camera.alt)} B=${Math.round(b.camera.alt)})`);
+  if (dHeading > 1) out.push(`heading Δ${dHeading.toFixed(1)}° (A=${a.camera.heading.toFixed(1)} B=${b.camera.heading.toFixed(1)})`);
+  if (dPitch > 1) out.push(`pitch Δ${dPitch.toFixed(1)}°`);
+  if (a.mode !== b.mode) out.push(`mode A=${a.mode} B=${b.mode}`);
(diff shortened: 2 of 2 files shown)


rdhyee:
## Retrospective Codex review (2026-05-11)

Posted on behalf of OpenAI Codex (gpt-5.4). Run after merge + deploy of #203 + #205. Codex read both diffs and surrounding context in `explorer.qmd` directly. Verbatim review below.

---

**Overall**: these PRs fixed the two observed round-trip failures in the right general direction, but the implementation is tactical and still leaves URL/state synchronization more fragile than I'd like. I would not revert it. I would file follow-ups.

### Correctness critique

The early URL write in the `camera.changed` debounce is mostly safe because it writes current viewer state with `replaceState`, and the trailing write later re-emits the final mode/state after async work settles. I do not see a direct ordering bug where the two corrupt each other.

The caveat: the early write captures the **current** mode, not the **intended target** mode. If the user crosses below `ENTER_POINT_ALT` and copies the URL while the res8 / point-mode transition is still pending, the URL can be "low altitude, no `mode=point`". That is better than the old stale camera URL, but it may still fail the intended "this view round-trips into point mode" contract because boot only explicitly forces point mode when the URL says `mode=point`.

`moveEnd` as added in #205 is idempotent with the `camera.changed` URL writes. Duplicate `replaceState` calls are cheap, and if `camera.changed` later transitions mode, its final write wins with the more complete state. I do not see a serious competition bug there.

The boot-end `tryEnterPointModeIfNeeded()` call is directionally correct. Running it after `_suppressHashWrite = false` is better than keeping URL writes suppressed during a potentially long cold-cache point-mode load. But there is a history-stack smell: `tryEnterPointModeIfNeeded()` eventually calls `enterPointMode()` with default `pushHistory`, so boot hydration can `pushState` during initial load. That behavior probably existed through the camera-handler path too, but the explicit boot path makes it more deterministic. Boot and hash hydration should generally reconcile state without adding a new browser history entry.

`_suppressHashWrite` covers the major intended paths: initial `setView` and hashchange-driven `flyTo`. The weak spot is that hashchange suppression is cleared by a fixed 2s timer, not by actual flight completion. Given the hashchange flight duration is 1.5s, this is probably okay in normal conditions, but it is still timer-coupled state management.

Programmatic `flyTo` calls outside hashchange are now URL-writing via `moveEnd`. That is usually desirable for search-result flights. But it also means any stale `selectedPid` / `selectedH3` already in `_globeState` will be preserved into the new camera URL unless that action explicitly clears or updates selection. That was already possible through `camera.changed`; #205 just makes it more reliable.

### Design critique

`moveEnd` is the right backstop for URL freshness. Lowering `percentageChanged` would increase churn and still would not give a clean "settled final camera" signal unless set very low. Cesium's docs define `changed` as threshold-based and `moveEnd` as the stopped-moving event, so `moveEnd` matches this problem better.

However, #205 only uses `moveEnd` for URL writes. That means sub-threshold moves still bypass the rest of the camera-settled pipeline: sample reloads, cached viewport re-rendering, and cluster count updates are still tied to `camera.changed`. If a small pan is important enough to update the URL, it may also be important enough to update "Samples in View" / "Clusters in View". I'd follow up by routing both `camera.changed` and `moveEnd` through one settled-camera reconciliation path, with cheap URL writes separated from expensive data work.

The early-write pattern is understandable but not a healthy long-term shape. There are now many URL writers: selection clicks, mode transitions, source/facet query writers, share button, camera debounce early/trailing writes, and `moveEnd`. I'd rather see a small explicit layer:

```js
writeGlobeHash({ replace: true })
reconcileCameraState({ reason, writeEarlyHash: true })
setExplorerMode(nextMode, { pushHistory: false | true })
```

Then boot, hashchange, user camera movement, and share/copy can call clear, intention-revealing paths.

For Bug B, I would not try to force Cesium to emit `camera.changed` from `setView`. The cleaner fix is what #203 did: after programmatic camera placement, call the app's own reconciliation logic explicitly. A tiny camera nudge would be hacky. `flyTo({ duration: 0 })` may give you a completion callback, but I would still not make app correctness depend on Cesium deciding to emit movement/change events for a zero-duration operation.

### What we missed

The harness tested camera + mode, but not the full URL contract.

Untested or suspicious cases:

- **`pid` + filters**: boot/hashchange PID hydration does not appear to apply `sourceFilterSQL()` or facet filters, so `?sources=...#...&pid=...` can hydrate a sample that the active filters exclude.
- **`h3` hashchange**: when `fetchClusterByH3(state.h3)` returns null, the hashchange path clears the UI but does not clear `viewer._globeState.selectedH3`. Boot does clear it. That asymmetry can cause invalid/filtered `h3` to be re-emitted later by `buildHash()`.
- `search`, `search_scope`, `sources`, `material`, `context`, `object_type`, and `view=table` are outside the round-trip harness.
- **Low-altitude URLs without `mode=point`** remain ambiguous. The early-write fix can create exactly that state while point-mode entry is pending.
- **Search-result flights** do not set or clear selection, so the URL can become "new camera, old selectedPid/h3".
- The "Samples in View = 5000" cap remains unresolved, as expected.

The mode-sync anomaly is worth filing. The code currently has two mode stores: closure-private `mode` and public `viewer._globeState.mode`. They are updated together today, but the duplication is fragile, and the harness/tests use `_globeState.mode` as truth. I would either make `_globeState.mode` the single source of truth or wrap mode transitions in one setter and add an invariant assertion/log.

The investigation script should be promoted, but not necessarily as-is. Keep the artifact, then extract CI-safe specs:

- point deep-link with omitted/default heading boots into point mode;
- sub-threshold pan updates hash after `moveEnd`;
- copied URL from context A restores camera/mode in context B;
- selection/filter URL cases for `pid`, `h3`, and `sources`.

The full cold-cache/live-site investigation can remain a manual diagnostic because it is slow and network-sensitive.

### Suggested follow-up issues

1. Fix #201 Part 1: make "Samples in View" honest when the point query hits `DEFAULT_POINT_BUDGET`; show capped state or compute a real count separately.
2. Centralize camera URL/mode reconciliation; make `moveEnd` drive settled-camera data/UI updates, not only URL writes.
3. Avoid `pushState` during boot hydration and hash hydration; mode reconciliation should support `{ pushHistory: false }`.
4. Define and test the low-altitude-without-`mode=point` contract.
5. Add Playwright specs for URL round-trip coverage beyond camera/mode: `pid`, `h3`, `search`, `search_scope`, `sources`, facets, and `view=table`.
6. Fix `h3` hashchange null-result handling so runtime state and URL stop preserving invalid/filtered H3 cells.
7. Investigate/remove the dual `mode` state (`mode` closure variable vs `viewer._globeState.mode`).
8. Clear or intentionally update selection state when search-result flights move the camera away from a selected sample/cluster.

rdhyee:
Follow-up issues filed from the Codex retrospective above, clustered for tractability:

- **#206** — "Samples in View" counter accuracy (= #201 Part 1, the user-visible 5000-cap bug). Starting here next.
- **#207** — URL state contract micro-fixes (low-altitude-without-mode=point, h3 null-result asymmetry, search-result flight selection, boot pushState avoidance). 4 small items, each independently shippable.
- **#208** — Refactor: centralize URL writers + collapse dual mode state. Bigger structural work; sequenced after #207 lands the targeted fixes.
- **#209** — Expand Playwright URL round-trip coverage to pid / h3 / search / scope / filters / view=table. Promotes the investigation harness to CI-safe specs.

Working through them in that order.

Issue #208, opened 2026-05-11
explorer: refactor URL/mode state management (centralize writers, collapse dual mode)


rdhyee:
Structural follow-up from Codex's retrospective on #203 + #205. Two related smells; addressing them together because they touch the same code paths.

## Smell 1: Many URL writers, no single boundary

Across `explorer.qmd` there are now (at least):
- Selection click handlers (sample, cluster) — `pushState`
- Mode-transition writes — inside `enterPointMode` / `exitPointMode`
- `writeQueryState()` — query-string filters
- Camera-debounce early write (#203)
- Camera-debounce trailing write (pre-existing)
- `camera.moveEnd` listener (#205)
- Share-button copy
- Hashchange handler's internal `replaceState`

Each write has its own gate (`_suppressHashWrite`, freshness checks, etc.). When something goes wrong with URL state, the answer to "which writer wrote that?" is non-obvious.

**Codex's sketched interface**:
```js
writeGlobeHash({ replace: true })
reconcileCameraState({ reason, writeEarlyHash: true })
setExplorerMode(nextMode, { pushHistory: false | true })
```

Then boot, hashchange, user camera movement, and share/copy call clear, intention-revealing paths.

Also flagged: `moveEnd` currently only writes the URL — sub-`percentageChanged` moves still bypass the rest of the camera-settled pipeline (sample reloads, cluster count refresh). If a small pan is important enough to update the URL, it should probably update the in-view stats too. A unified "settled-camera reconciliation" path would solve both.

## Smell 2: Dual `mode` state

`explorer.qmd` carries the current mode in two places:
- Closure-private `mode` variable (used by the camera-changed handler at lines 1971+)
- Public `viewer._globeState.mode` (used by `buildHash`, tests, harnesses)

They're updated together today but the duplication is fragile. The Playwright investigation harness's iter 1 captured `viewer._globeState.mode === 'cluster'` shortly after `waitReadyAndPointMode` saw `mode === 'point'` — empirical evidence of transient sync mismatch.

**Fix shape**: make `_globeState.mode` the single source of truth, and wrap mode transitions in one setter (`setExplorerMode(next, opts)`). Add an invariant assertion / dev-time console log when the two diverge until the closure variable can be deleted.

## Sequencing

Probably:
1. Collapse `mode` to single source (smell 2) — smaller, sets the foundation.
2. Introduce `writeGlobeHash` + `setExplorerMode` and migrate the writers (smell 1).
[...]

PR #210, opened 2026-05-11
explorer: report real Samples-in-View count when cap is reached (closes #206)

### LINKED ISSUE #206: explorer: 'Samples in View' counter is the fetch budget, not the real count (#201 Part 1)
Originally filed as Part 1 of #201. Splitting out as a dedicated issue since #201 was closed by #203 / #205 (which fixed Part 2 only).

## Symptom

The "Samples in View" stat box reads exactly 5,000 in dense regions — the value of `DEFAULT_POINT_BUDGET` at `explorer.qmd:418`. In Cyprus (lat ≈ 34.99, lng ≈ 33.70), direct DuckDB query against `data.isamples.org/isamples_202601_samples_map_lite.parquet` returns 23,421 samples in a ±0.1° box. The counter underreports by ~5x there. The cluster is one dense site (almost certainly Polis Excavations, OPENCONTEXT source).

## Root cause

```sql
  ${sourceFilterSQL('source')}
  ${facetFilterSQL()}
LIMIT 5000
```

`explorer.qmd:1557`:

```js
```

`cachedData.length` IS the row count of the LIMIT 5000 result. The counter therefore tops out at 5000 by construction.

Secondary smells:
- **No `ORDER BY` before `LIMIT`** → which 5000 rows return is undefined (probably stable in DuckDB-on-parquet but not contractual).
- **Label says "in View" but fetch uses a padded (30%) viewport** (`explorer.qmd:1514-1522`). Even ignoring the cap, the count meaning is loose.
- `renderSamplePoints` plots all of `cachedData` including rows outside the actual viewport.

## Fix directions (from Codex retrospective on #203)

In rough order of effort:

1. **Honest relabel** (cheapest): change the label to "Samples Loaded (max N)" and wire the budget value into the label. Counter stops lying.
2. **Compute real count alongside**: a fast `SELECT count(*)` against the same WHERE (no LIMIT) is cheap on the lite parquet via DuckDB-WASM range reads. Display "X loaded / Y total in view", with explicit signaling when Y > X.
3. **Adaptive aggregation**: if real count > budget, fall back to a cluster-style representation or surface a "too dense to render individually — Y samples here" affordance.
4. **Add `ORDER BY pid`** to the point query so the 5000 subset is at least deterministic across browsers and sessions.

Direction 2 (real-count alongside) is probably the right user-visible answer; direction 4 is independent and could ship with any of the others.

## Acceptance

- [ ] Counter accurately represents the in-view sample count, or is unambiguously labeled as a capped/loaded count.
- [ ] Cyprus deep-link (`#v=1&lat=34.9957&lng=33.6798&alt=15212&mode=point`) shows a number that does not silently understate the real density.
- [ ] No regression in cluster-mode "Samples in View" (which already counts viewport intersections correctly).

rdhyee:
Closes #206. First item from the #203 retrospective Codex review (cluster D).

## Problem

In point mode, the viewport sample query is capped at `POINT_BUDGET = 5000` with no `ORDER BY`. The "Samples in View" stat box displayed `cachedData.length`, which equals the cap in any region dense enough to exceed it. Cyprus deep-link (`#v=1&lat=34.9957&lng=33.6798&alt=15212&mode=point`) showed exactly 5,000 — real count from a direct DuckDB query against the lite parquet: **23,421** samples in the same viewport. ~5x understatement.

## Fix

Two changes in `loadViewportSamples`:

1. **Real count**: after the LIMIT POINT_BUDGET data fetch, run a `count(*)` against the same WHERE clause **only when the cap is reached**. Stat box `sSamples` now displays the true count; `sPoints` shows the rendered count under a "Samples Rendered" label so the user sees both numbers. Phase message becomes "23,422 samples in view (showing 5,000 — zoom in for more). Click one for details." when capped. Stale-guard applied between the two queries.

   Below the cap the second query is skipped (`data.length == real count` by definition), so no extra round-trip in the common case.

2. **`ORDER BY pid`** on the data query, so the POINT_BUDGET-worth of rows the LIMIT returns is deterministic across browsers and sessions. (Codex recommendation 4 from the #203 retrospective.)

## Verification

Smoke test against localhost with this branch (Quarto preview):

```
Cyprus (alt=15212) point-mode display:
  mode: point
  Samples Rendered: 5,000
  Samples in View: 23,422
  phase: 23,422 samples in view (showing 5,000 — zoom in for more). Click one for details.
```

The 23,422 real count matches the OC investigation's DuckDB result (23,421 at ±0.1°; 23,481 at ±0.13° — explorer's 30%-padded viewport sits between).

`url_roundtrip_investigation.js` still passes all 6 iterations against localhost, so no regression in #203/#205 URL state work.

## Risks

- **One extra DuckDB-WASM query per point-mode load *when capped only*.** count(*) on a filtered parquet scan is comparable cost to the data fetch but returns a single number — under 100ms on warm cache in typical viewports. Acceptable for the UX win, especially since dense viewports are exactly the case where users were getting wrong information.
- **`ORDER BY pid`** adds a sort step. Affects performance only in the capped case (sort happens on the result of the WHERE-filtered scan). pid is the primary key on the parquet — sort is on small string keys.
- **Stat box label change** ("Samples Rendered" appears when capped). Could be jarring if a label-stability test exists; none in the current spec set.

## Test plan

- [x] Smoke: Cyprus deep-link shows real count = 23,422, rendered = 5,000
- [x] url_roundtrip_investigation.js — all 6 iterations green against localhost
- [ ] Manual smoke on deploy preview: Cyprus + a non-capped viewport (small island somewhere with <5k samples)
- [ ] Codex review (recommended)

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -1527,14 +1527,22 @@ zoomWatcher = {
             performance.mark('sp-s');
             // facetFilterSQL() returns a portable `pid IN (...)` predicate,
             // so the same query works whether or not facet filters are active.
-            const query = `
-                SELECT pid, label, source, latitude, longitude,
-                       place_name, result_time
-                FROM read_parquet('${lite_url}')
+            //
+            // ORDER BY pid (issue #206): without explicit ordering, which
+            // POINT_BUDGET-worth of rows the LIMIT returns is undefined and
+            // can differ across browsers/sessions for the same query.
+            const whereClause = `
                 WHERE latitude BETWEEN ${padded.south} AND ${padded.north}
                   AND longitude BETWEEN ${padded.west} AND ${padded.east}
                   ${sourceFilterSQL('source')}
                   ${facetFilterSQL()}
+            `;
+            const query = `
+                SELECT pid, label, source, latitude, longitude,
+                       place_name, result_time
+                FROM read_parquet('${lite_url}')
+                ${whereClause}
+                ORDER BY pid
                 LIMIT ${POINT_BUDGET}
             `;
             const data = await db.query(query);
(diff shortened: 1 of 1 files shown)


rdhyee:
## Codex review applied (commit ca8407c)

Posted the Codex review on this PR. Acted on it. Three changes:

1. **Stale guard in count-query catch**. The Codex-flagged race (count fails after a newer request starts → stale catch path mutates state) is closed: `if (myReqId !== requestId) return` at the top of the catch.

2. **Cache-hit stats**. `cachedTotalCount` + `cachedCapReached` now cached alongside `cachedBounds`/`cachedData`. `loadViewportSamples`'s cache short-circuit now re-applies them to the stat boxes, so panning inside the padded cache window doesn't leave stale figures on screen. All four `cachedBounds = null` reset sites also clear the new fields.

3. **Always-on stable labels**. Dropped the conditional label flip (`Samples in View` ⇄ `Samples Rendered` as cap is crossed) in favor of stable labels: sPoints always says `Samples Rendered`, sSamples always says `Samples in View`. Redundancy when uncapped is intentional per Codex's preference.

Filed Codex's other suggestion as **#211**: tighten the count to the visible viewport rather than the 30%-padded fetch box. Out of scope here; folds naturally into #208 if scope grows.

Verified locally:
- Cyprus smoke still reads `Samples Rendered: 5,000 / Samples in View: 23,422`
- `url_roundtrip_investigation.js` all 6 iterations green

Codex review verbatim (for the record):

> 1. **Race bug in count-query failure path** [...] If the count query fails after a newer camera/filter request starts, the stale request can still update cachedBounds, cachedData, points, stats, and phase text. Add the same guard inside the catch before logging/fallback.
> 2. **Cache-hit stats remain stale** [...] Panning inside the padded cached area can leave the previous 'Samples in View' count on screen.
> 3. **'Samples in View' is still padded-area count** [...] If the goal is strict honesty, count against actual bounds; if the goal is only 'not silently capped,' this is acceptable but should probably be called out as a follow-up. [→ #211]
> 
> UX: I would prefer always-stable labels: `Samples Rendered` and `Samples in View`, even when equal. [✅]
> 
> I would not defer this to #208. Add the stale guard in the count catch, then ship. [✅]

Ready to merge from my side.

Issue #209, opened 2026-05-11
explorer: expand Playwright URL round-trip coverage beyond camera+mode


rdhyee:
Codex retrospective on #203 + #205 noted the URL round-trip harness (`tests/playwright/url_roundtrip_investigation.js`) only covers `camera` + `mode`. The URL state contract is larger.

## Untested cases

From Codex's review:

- [ ] **`pid` deep-link + active filters**: boot/hashchange PID hydration does not appear to apply `sourceFilterSQL()` or facet filters, so `?sources=...#...&pid=...` can hydrate a sample that the active filters exclude. Need a round-trip test.
- [ ] **`h3` deep-link round-trip**: open `#h3=...` in a fresh tab; cluster card hydrates; URL persists across pan/zoom.
- [ ] **`search` + `search_scope`**: `?search=pottery&search_scope=area` should restore the search input, scope toggle, and re-run.
- [ ] **`sources`, `material`, `context`, `object_type`** facet filters: each should round-trip via `?sources=`, etc.
- [ ] **`view=table`**: table mode should restore.
- [ ] **Search-result flight**: clicking a search result flies the camera; URL captures the new camera and either reflects or clears prior selection. Confirm round-trip.
- [ ] **Sub-threshold pan settles via `moveEnd`** (regression for #205): a small pan with no zoom should update the URL within ~1s of mouseup.
- [ ] **Cold-cache point-mode deep-link**: with `?cache=disabled` or a wiped browser cache, the deep-link should still round-trip — measure time-to-stable, assert eventual consistency.

## Harness vs spec

The existing `url_roundtrip_investigation.js` is a Node script useful for manual investigation but doesn't run in CI. Plan:

1. Keep the investigation harness as a manual diagnostic (it does cold-cache live-site work that's too slow / network-sensitive for CI).
2. Extract CI-safe Playwright specs into `tests/playwright/url_roundtrip.spec.js` covering the items above, hitting `localhost:5860` like the other specs.
3. Each spec asserts: open URL in fresh context, wait for settle, snapshot relevant state, compare against an expectation built from the source URL.

## Acceptance

- New `tests/playwright/url_roundtrip.spec.js` (or split into focused spec files) running in CI.
- All items above pass against `localhost:5860` (Quarto preview of current main).
- Either the existing investigation harness is updated to share helpers with the specs, or it's left as-is with a comment pointing to the specs.

## Refs

- PR #203 — the original URL round-trip fix
[...]

PR #212, opened 2026-05-12
explorer: URL state contract micro-fixes (closes #207)

### LINKED ISSUE #207: explorer: URL state contract micro-fixes (post-#203 follow-ups)
Cluster of small, targeted URL-state contract fixes surfaced by Codex's retrospective review of #203 + #205. Each item is independently shippable; bundling for tracking, not for one PR.

## Items

- [ ] **Low-altitude-without-`mode=point` round-trip** (Codex item 4). The early-write fix in #203 can produce URLs like `alt=8000` *without* `mode=point` if the user copies during a pending point-mode transition. Boot only force-enters point mode when the URL says `mode=point`, so the round-trip lands in cluster mode at low altitude — not what the user saw. **Fix shape**: boot path should enter point mode based on altitude (alt < `ENTER_POINT_ALT`) AND/OR the URL flag, not the URL flag alone. Update `tryEnterPointModeIfNeeded()` call at boot end to fire whenever altitude is below threshold, regardless of `ih.mode`.

- [ ] **`h3` hashchange null-result asymmetry** (Codex item 6). When `fetchClusterByH3(state.h3)` returns null in the hashchange handler (`explorer.qmd:~2568-2580`), the UI is cleared but `viewer._globeState.selectedH3` is NOT cleared. Boot does clear it (line ~2590). The asymmetry means an invalid/filtered `h3` can be re-emitted later by `buildHash()`. **Fix shape**: clear `viewer._globeState.selectedH3` in the hashchange null branch to match the boot behavior.

- [ ] **Selection state with search-result flights** (Codex item 8). Search-result clicks fly the camera but don't set or clear `selectedPid` / `selectedH3`. The URL then becomes "new camera, old selection". **Fix shape**: search-result click handler should either set selection to the searched-for sample or clear stale selection before flying.

- [ ] **Avoid `pushState` during boot/hash hydration** (Codex item 3). `tryEnterPointModeIfNeeded()` at boot eventually calls `enterPointMode()` with default `pushHistory`, so boot hydration can add a browser history entry. Boot and hashchange should reconcile state without growing history. **Fix shape**: thread an explicit `{ pushHistory: false }` option through the boot call site.

## Why bundled

Each is a small, localized fix in `explorer.qmd`. None unblocks the others. Codex flagged them as a group; tracking them together keeps the URL state contract's edges visible while we work through them. Individual PRs are fine.

## Refs

- #201 / #203 / #205 — original URL round-trip work
- PR #203's retrospective Codex review comment — full context
- #164 — Explorer state contract (these items belong in that inventory once stabilized)

rdhyee:
Closes #207. Four small, independent fixes from the #203 retrospective Codex review (cluster A).

### #207 item 4 — low-altitude-without-`mode=point` round-trip

**Problem**: #203's early URL write can produce URLs like `alt=8000` *without* `mode=point` if the user copies during a pending point-mode transition. Boot only force-entered point mode when the URL had `mode=point`, so the round-trip landed in cluster mode at low altitude.

**Fix**: boot also enters point mode when `ih.alt < ENTER_POINT_ALT`, regardless of `ih.mode`. `tryEnterPointModeIfNeeded()` already short-circuits above the threshold, so cluster deep-links are unaffected.

### #207 item 6 — `h3` hashchange null-result asymmetry

**Problem**: when `fetchClusterByH3(state.h3)` returned null in the hashchange handler, the UI was cleared but `viewer._globeState.selectedH3` was left set to the invalid value. Boot path already cleared it on the same condition.

**Fix**: hashchange null branch now sets `selectedH3 = null` to match boot, preventing `buildHash()` from re-emitting invalid H3.

### #207 item 8 — selection state with search-result flights

**Problem**: search-result row clicks flew the camera but didn't update selection. The URL captured at flight settle (via #205's `moveEnd` listener) carried whatever prior `selectedPid` / `selectedH3` was set, misrepresenting the user's actual view.

**Fix**: 
- Row click now sets `selectedPid` to the clicked sample's pid (treats click as an explicit selection, matching on-globe sample clicks).
- Auto-flight to first result on world-scope searches clears prior selection (treats auto-pan as a nudge, not an explicit selection — user clicks a specific row to establish a new selection).

### #207 item 3 — boot/hash hydration shouldn't grow history

**Problem**: `tryEnterPointModeIfNeeded()` called `enterPointMode()` with default `pushHistory`, so boot would `pushState` a history entry every time.

**Fix**: threaded `opts.pushHistory` through the helper; boot call site passes `{ pushHistory: false }`. Camera-handler call sites keep the default (`pushState`) since those are user-driven movements where adding history is intentional.

## Verification

- `tests/playwright/url_roundtrip_investigation.js` — all 6 iterations green on localhost
- One-off test: URL `#v=1&lat=35.0150&lng=33.7000&alt=8000` (low alt, NO mode param) now enters point mode in a fresh tab and shows the real 23k count (combination of #206 + #207 item 4).

## Risks

- **Item 4 broadens boot's point-mode entry condition.** Previously only `ih.mode === 'point'` triggered the explicit boot call; now altitude does too. Cluster deep-links above `ENTER_POINT_ALT` are unchanged (short-circuit). A bookmarked URL with stale `mode=cluster` and low altitude will now enter point mode at boot — but that's the correct behavior given the altitude is the more authoritative state.
- **Item 8 row-click selection.** Setting `selectedPid` from a search-result click is a small UX change: the URL now reflects the searched-for sample as selected. The clicked sample's full sample card hydration is not part of this PR — that requires more data than the search-result row carries. Tracked as a possible follow-up if the lack of side-panel hydration is jarring.
- **Item 3 history avoidance.** Boot no longer adds a history entry on point-mode entry. Users navigating away and using back will now go to the previous page rather than a "stuck on this page in cluster mode" intermediate state. Net win, no obvious regression.
- **Tests don't cover items 6 and 8 directly.** They're tracked in #209 (test coverage expansion). The fixes are small enough that code review covers them.

## Test plan

- [x] url_roundtrip_investigation.js all 6 green
- [x] Low-alt no-mode URL enters point mode (verified manually)
- [ ] Manual smoke on deploy preview: 
  - Cyprus deep-link → pan small → copy URL → paste → matches (regression check for #203 + #205)
  - URL with low alt but no `mode=point` → enters point mode (item 4)
  - Hash to invalid h3 → URL cleans up (item 6)
  - Search result click → URL reflects pid (item 8)
- [ ] Codex review (recommended)

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -2226,12 +2226,17 @@ zoomWatcher = {
             duration: 1.5,
         });
 
-        // After flight settles, force mode and clear suppress flag
+        // After flight settles, force mode and clear suppress flag.
+        // Use the same altitude-authoritative predicate as the boot path
+        // (issue #207 item 4 → Codex follow-up): without this, back/forward
+        // through a `#alt=8000` URL with no `mode=point` would exit point
+        // mode here even though boot would have entered it.
         viewer._suppressTimer = setTimeout(() => {
             viewer._suppressHashWrite = false;
             const s = readHash();
-            if (s.mode === 'point' && mode !== 'point') enterPointMode(false);
-            else if (s.mode !== 'point' && mode === 'point') exitPointMode(false);
+            const wantsPoint = s.mode === 'point' || (s.alt != null && s.alt < ENTER_POINT_ALT);
+            if (wantsPoint && mode !== 'point') enterPointMode(false);
+            else if (!wantsPoint && mode === 'point') exitPointMode(false);
         }, 2000);
 
         // Handle pid / h3 selection (sample mode wins if both present — see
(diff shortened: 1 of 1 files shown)


rdhyee:
## Codex review applied (commit 5176967)

Codex's review surfaced two real issues I'd missed; both now addressed.

**1. Hashchange handler parity with boot's altitude-authoritative predicate** (item 4 follow-up). Boot now treats `alt < ENTER_POINT_ALT` as authoritative for entering point mode, but the hashchange post-flight setTimeout at `explorer.qmd:2233` was still gated on `s.mode === 'point'` only. Result: back/forward through a `#alt=8000` URL with no `mode=point` would *re-enter cluster mode* via the timeout, even though boot would have entered point mode for the same URL. Fixed by using the same `wantsPoint = s.mode === 'point' || (s.alt != null && s.alt < ENTER_POINT_ALT)` predicate in both places.

**2. Search-row click now does the full sample-selection ceremony** (item 8 follow-up). My first cut just set `selectedPid` + flew. Codex correctly pushed back: without freshness-token bump, sample-card hydration, and lazy detail load, a slow prior selection's repaint could land *after* the click, leaving UI inconsistent with `_globeState` and the URL. Now matches the on-globe sample click at line ~944:
- `freshSelectionToken(viewer)` bump
- `updateSampleCard` with full metadata from the search result row (built a `Map<pid, result>` for closure capture)
- flyTo
- Lazy detail query with stale guards in both success and catch paths

samplesSection is intentionally NOT cleared (unlike the globe click handler) — it holds the search results the user is browsing.

Codex confirmed items 3 and 6 are correctly handled (three-state `opts?.pushHistory` is right; null-h3 clear is race-safe via the `isStale()` check before mutation).

Verified: `url_roundtrip_investigation.js` all 6 iterations still green.

Codex review verbatim:
> 1. `explorer.qmd:2233` still makes hashchange mode reconciliation depend only on `s.mode === 'point'`. So boot now treats `alt < ENTER_POINT_ALT` as authoritative, but back/forward or manual hash edits do not. [...] If altitude is authoritative, use the same predicate here as boot. [✅]
> 2. `explorer.qmd:2515` and `explorer.qmd:2538` mutate selection state without treating search as a real selection event. [...] A slow previous globe/hash selection can still repaint the panel after the search click [...] I would not leave this as a follow-up if this PR is about URL selection truth. [✅]
> No other blocking issues from the diff.

Ready to merge from my side.

Issue #234, opened 2026-05-23
explorer: architectural direction — make filter semantics coherent across all surfaces


rdhyee:
## Purpose

Capture the architectural direction for the explorer's filter semantics, sequenced as a roadmap, with explicit decision rationale. Filed for review (Codex, Gemini, and human collaborators). The goal is alignment *before* implementation begins on the substantive steps.

## Context

The explorer has five surfaces that show numbers about samples — map dots, "Samples in View" stat, "Samples Rendered" stat, samples table count, facet-legend counts, and the search-results line. Today each surface applies a different combination of constraints:

| Constraint | Map / Stats / Table | Facet counts | Search results line |
|---|---|---|---|
| Source checkboxes | ✅ | ✅ | ✅ |
| Material / Sampled Feature / Specimen Type | ✅ | ✅ | ✅ |
| Bbox (viewport) | ✅ | ❌ | optional (area-scope) |
| Search text | ❌ | ❌ | ✅ |

Result: three different "filter" semantics on one page. The 2026-05-22 investigation session (see `#229` closure note, and the design briefing in `~/dev-journal/projects/isamples-facets.md`) hit three concrete confusions stemming from this:

- "I have `pottery Cyprus` in the search box but the facet counts and 'Samples in View' don't reflect it."
- "I filtered to `material=soil` but the cluster dots include non-SESAR colors even though most soil is SESAR."
- "What does '5,451 samples match the current filters' actually mean?"

## The decision space

Three orthogonal axes (named for cross-reference):

- **A1**: search is a global filter — restricts map, table, stats, facet counts.
  **A2**: search is a side-panel lookup — restricts only the search-results list. *(current)*
- **B1**: facet counts reflect viewport bbox — pan, counts change.
  **B2**: facet counts stay global regardless of viewport. *(current)*
- **C1**: cluster mode honestly reflects facet filter — H3 dots per filtered subset *(expensive — pre-bake per facet, or live aggregate)*.
  **C2**: cluster mode ignores filter, surface this loudly with `#facetNote`. *(current — but the note is bugged on URL-load)*.
  **C3**: auto-switch to point mode when any facet is active — no cluster dishonesty, but point density problems (see #231).

## Direction picked

[...]

PR #236, opened 2026-05-23
explorer: search-results "50+" → real "N of M" count (closes #232)

### LINKED ISSUE #206: explorer: 'Samples in View' counter is the fetch budget, not the real count (#201 Part 1)

### LINKED ISSUE #232: explorer: search-results '50+' label should show the real count
## Symptom

The search-results line under the in-map search box reads `50+ results for "pottery Cyprus"`. The `+` is appended whenever the result set hits the hardcoded `LIMIT 50`, hiding the actual count — which could be hundreds or thousands.

For `pottery Cyprus` in the world (no area scope), there are ~7,124 matching samples (per the comment at `explorer.qmd` L3229). The user sees `50+`. Significant under-disclosure.

## Code location

- `explorer.qmd` L3296, L3313: `LIMIT 50` on both the area-scope and world-scope queries
- `explorer.qmd` L3327:
  ```js
  searchResults.textContent = `${results.length}${results.length === 50 ? '+' : ''} results for "${term}"`;
  ```

## Suggested fix

Mirror the pattern already used by `loadViewportSamples()` at L2395-2415: when the LIMIT is hit, run a separate `COUNT(*)` query (same WHERE clause, no ORDER BY, no LIMIT) and display the real total.

```sql
SELECT COUNT(*) AS n
FROM read_parquet('${facets_url}') f
[INNER JOIN ... if area scope]
WHERE ${searchWhere}
  [${bboxSQL} if area scope]
  ${sourceFilterSQL('f.source')}
  ${facetFilterSQL()}
```

Display becomes: `50 of 7,124 results for "pottery Cyprus"` (or similar — Hana to weigh in on copy if needed).

## Considerations

- Cost: an extra COUNT scan against the facets parquet. For broad terms this is the same scan the LIMIT query already does — so it should not double the wall-clock time, just add some overhead.
- Cancellation: the search has a freshness token at `searchSeq`/`searchAbortController`; the COUNT query needs to honor it too.
- Only run COUNT when `results.length === 50`. Skip it when fewer rows returned — that count *is* the truth.

## Why this matters

Without a real count, users can't tell whether `50+` means 51 or 50,000. That affects whether they refine their search or just look elsewhere. It's also a small piece of the larger trust gap around "what does the search actually do?" (related to the global-filter vs side-panel decision in #164).

## Out of scope

- Whether search should restrict the map / table / facet counts (covered by #164 / #230).
- Pagination of search results past 50 (would be a separate feature).
- Replacing the LIMIT itself with a higher cap (low value vs. just showing the real count).

## Acceptance

- Search line reads e.g. `50 of N results for "term"` when the cap is hit.
- Search line reads e.g. `12 results for "term"` when below the cap (unchanged from today).
- No regression in cancellation behavior when the user types quickly.

## Cross-refs

- #164 (search-semantics decision — broader context)
- #168 (the comment block above the search query notes the interim recall fix)
- #230 (remaining facet-count work)

rdhyee:
## Summary

Roadmap **step 2** of #234 (#232 "50+" → real count, quick-win).

The in-map search-results line used to read `50+ results for "pottery"` whenever the SELECT hit `LIMIT 50`, hiding whether the true result set is 51 or 50,000. For `pottery Cyprus` (world scope) the real count is ~7,124; users saw `50+` and couldn't gauge whether to refine.

**Before:** `50+ results for "pottery"`  
**After:** `50 of 7,124 results for "pottery"`

Mirrors the pattern used to close #206 (real-count follow-up in `loadViewportSamples()`).

## Approach

- When the SELECT cap is hit, fire a follow-up `COUNT(*)` with the same WHERE clause and replace the label.
- Below the cap, `results.length` IS the truth — no second scan.
- Initial render keeps `50+` while COUNT is in flight, so users see "more exist" signaling immediately.

## Key implementation details

1. **Effective query shape.** The area-scope path can fall back to the world-mode SELECT when `viewerBboxSQL()` returns null. Track which shape actually ran (`effectiveQueryShape` / `effectiveBboxSQL`) so the COUNT uses the matching WHERE/JOIN. Keying off `effectiveScope === 'area'` would have generated a malformed COUNT in the fallback case.

2. **Cancellation.** A second search bumps `_searchSeq`. The COUNT guards against `searchId !== _searchSeq` so a stale COUNT can't overwrite a fresher search's text or pollute its structured log.

3. **Failure handling.** A COUNT error leaves the initial `50+` label in place — under-disclosure beats clearing the line. Logged via `console.warn`; `count_ms` is still recorded.

4. **Structured log (#167) additions.** `total_count` (real count, or `null` when not computed) and `count_ms` (COUNT scan duration). `elapsed_ms` continues to mark end-of-handler so it remains an end-to-end "search complete" latency including the COUNT.

## Test plan

- [x] New spec `tests/playwright/search-real-count.spec.js`:
  - cap-hit case (`pottery`): initial `50+` then `50 of N (N>50)`, structured log carries `total_count` and `count_ms > 0`.
  - no-results case (`xyzzyqqqplugh`): still short-circuits to `No results`; COUNT not fired; `total_count: null`, `count_ms: 0`.
- [x] **Fault-injection**: both new tests FAIL against pre-fix code, PASS against the fix.
- [x] Local regression: `search-real-count` + `url-roundtrip` + `facetnote-url-load` → 9 passed (1.7m).
- [ ] Manual: verify against the cold-cache deploy that the COUNT scan time is acceptable for `pottery` / `pottery Cyprus`.

## What's NOT in this PR

- Search-as-global-filter (A1) — that's step 5 of #234, gated on FTS landing first.
- Pagination past 50 — out of scope per #232.
- Real BM25/FTS substrate — #169-#172 line of work.

## Acceptance criteria (from #232)

- [x] Search line reads `50 of N results for "term"` when the cap is hit.
- [x] Search line reads `12 results for "term"` (unchanged) when below the cap.
- [x] No regression in cancellation behavior when the user types quickly (guarded via `_searchSeq` invariant; covered conceptually rather than via a flaky two-SELECT race in Playwright).

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -3217,6 +3217,8 @@ zoomWatcher = {
         const tStart = performance.now();
         const terms = searchTerms(term);
         let resultsCount = 0;
+        let totalCount = null;     // Real count when cap is hit (#232); null otherwise.
+        let countMs = 0;           // Duration of the follow-up COUNT scan, ms.
         let errorMessage = null;
 
         try {
@@ -3272,7 +3274,14 @@ zoomWatcher = {
             // because area-scoped search by definition requires
             // coordinates. Drop coord-less samples before ranking; apply
             // the viewport predicate; THEN top-50.
+            //
+            // Track which shape actually ran so the follow-up COUNT(*)
+            // query for #232 uses the matching WHERE/JOIN. The area-scope
+            // path falls back to `world` when `viewerBboxSQL` returns null,
+            // so `effectiveScope === 'area'` is NOT a reliable predicate.
             let results;
+            let effectiveQueryShape = 'world';
+            let effectiveBboxSQL = null;
             if (effectiveScope === 'area') {
                 const bboxSQL = viewerBboxSQL('l.latitude', 'l.longitude');
                 if (!bboxSQL) {
@@ -3295,6 +3304,8 @@ zoomWatcher = {
                         ORDER BY relevance_score DESC, f.label
                         LIMIT 50
                     `);
+                    effectiveQueryShape = 'area';
+                    effectiveBboxSQL = bboxSQL;
                 }
             } else {
                 results = await runWorldQuery();
@@ -3324,7 +3335,18 @@ zoomWatcher = {
                 searchResults.textContent = `No results for "${term}"`;
                 return;
             }
-            searchResults.textContent = `${results.length}${results.length === 50 ? '+' : ''} results for "${term}"`;
+
+            // Initial render. The "+" indicates "more exist" when the cap was
+            // hit; if the cap was hit, the follow-up COUNT below replaces
+            // this with the real total ("50 of 7,124 results"). This keeps
+            // the user informed during cold-cache COUNT scans (where the
+            // second query can add a few hundred ms) rather than leaving the
+            // bare result-row count visible while the precise total
+            // resolves. See issue #232.
+            const capReached = results.length === 50;
+            searchResults.textContent = capReached
+                ? `${results.length}+ results for "${term}"`
+                : `${results.length} results for "${term}"`;
 
             // Show results in the samples panel
             const sampEl = document.getElementById('samplesSection');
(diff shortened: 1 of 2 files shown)


rdhyee:
## Codex review iteration log (4 rounds → LGTM)

| Round | Codex finding | Resolution | Commit |
|---|---|---|---|
| 1 | Stale `_searchSeq` guard placement: UI mutations + COUNT launch could fire for superseded searches before guard | Added early `searchId !== _searchSeq` check immediately after SELECT, before any UI write / panel render / fly-to / COUNT | `27aa70f` |
| 1 | Predicate not snapshotted: filter toggle mid-flight produces "50 of M" where M is for a different filter than the rendered 50 | Snapshot `sourceSQL` / `facetSQL` once at search-fire time, reuse in SELECT (both shapes) and COUNT (both shapes) | `27aa70f` |
| 2 | Stale error UI write: `catch` block unconditionally overwrites live UI with `Search error: …` if a superseded SELECT later rejects | Guard the catch's `searchResults.textContent` write on `searchId !== _searchSeq` (sets `wasSuperseded = true` and logs the original error without UI mutation) | `764ed63` |
| 2 | DOM-live telemetry in `finally`: `has_source_filter` / `has_facet_filter` could reflect post-toggle state | Snapshot `hadSourceFilter` / `hadFacetFilter` at function scope at search-fire time, reused in `finally` | `764ed63` |
| 3 | Invalid-submit stale overwrite: empty/short term early-return didn't bump `_searchSeq`, so a prior search's rejection could overwrite "Type at least 2 characters" | Bump `_searchSeq` at the TOP of `doSearch`, before any guard — every submission supersedes in-flight work | `5de9136` |
| 3 | Race test timing-dependent: wait accepted either `50+` or final `50 of N`, so the toggle could miss the window | Monkey-patch `db.query` with a 1.5s delay on `COUNT(*)` queries; wait for EXACT `50+` before toggling | `5de9136` |

**Round 4: LGTM** (Codex)

> The COUNT-delay test closes the round-3 coverage hole in practice: the patch is installed before the search, the test waits for exact `50+`, and COUNT is held open before the final log. […] Round 4 verdict: LGTM.

Known follow-up filed as a TODO comment in the test:

- Independent `has_source_filter` snapshot coverage requires a non-Quarto fixture — Quarto's see-also rendering duplicates `#sourceFilter` checkboxes, so toggle-and-count assertions are ambiguous in preview. The production snapshot uses the same pre-await pattern as `has_facet_filter`, which IS covered.

Final test count: **8 passed** (3 search-real-count + 5 url-roundtrip + 2 facetnote-url-load, total ~2 min) plus the fault-injection sanity check on every round.

Thanks codex for the four-round audit — the catch-path supersession (round 2) and invalid-submit window (round 3) were both real bugs I'd have missed.

PR #235, opened 2026-05-23
explorer: sync #facetNote on URL deep-link + mode transitions (#234 step 1)

rdhyee:
## Summary

Roadmap **step 1** of #234 (filter-semantics architectural direction). Pure bug fix, no behavior change beyond making the cluster-mode honesty note (`#facetNote`) honest in two cases where it had drifted out of sync.

- **URL deep-link**: arriving at `/explorer.html?material=…` restored checkbox state via `setCheckedValues`, which doesn't fire the `change` event, so `#facetNote` stayed hidden even with active filters. A user opening a shared URL saw dimmed cluster dots with no explanation.
- **Mode transitions**: the note was only updated inside `handleFacetFilterChange`. A user in cluster mode with the note showing kept seeing it after zooming to point mode (and vice versa).

Fix: extract `syncFacetNote()` as the single source of truth for the visibility invariant

```
visible ⇔ (any facet active) ∧ (mode === 'cluster')
```

and call it from the four state-mutation sites:

- `facetFilters` cell (URL deep-link restore)
- `handleFacetFilterChange` (user toggles a facet checkbox)
- `enterPointMode` / `exitPointMode` (mode transitions)

## Test plan

- [x] New spec `tests/playwright/facetnote-url-load.spec.js`:
  - cluster mode + `?material=<uri>` → `#facetNote` visible
  - cluster mode + no facet params → `#facetNote` hidden (negative control)
- [x] **Fault-injection**: confirmed the new spec FAILS against pre-fix code (`expected: 'block', received: 'none'`) and PASSES against the fix.
- [x] Full `tests/playwright/url-roundtrip.spec.js` (5 tests, includes point-mode deep-link and cross-context round-trip) passes — no regression.
- [ ] Manual: open `/explorer.html?material=<any URI>` and confirm note appears under the cluster legend.

## Notes for review

This PR is intentionally scoped to step 1 of the #234 roadmap (the 1–2 hour bug fix). It does **not** touch:

- `applyQueryToFacetFilters` itself (the URL-param read path is unchanged)
- The `refreshFacetCounts` cross-filter live-query path (B1 work, step 3)
- The cluster-mode H3-summary-vs-filter honesty story (C3 work, step 8)

The roadmap synthesis with sign-off and revised sequencing is in issue #234.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #230, opened 2026-05-22
explorer: facet counts — remaining work after #157/#226


rdhyee:
## Context

Cross-filter facet counts are already shipped (commit 7cee6a41, 2026-05-01).
The current implementation:

- Cube fast-path for single-axis-single-value selections (reads
  `isamples_202601_facet_cross_filter.parquet`, 526 cells)
- Live DuckDB-WASM `GROUP BY` against `isamples_202601_sample_facets_v2.parquet`
  (5.98M rows) for multi-axis and multi-select-within-axis cases
- 250 ms debounce, `facetCountsReqId` cancellation token, `.recomputing` italic
  loading state, `.facet-row.zero` dimming when count = 0

See `explorer.qmd` lines 2589-2705 + 977-997 for the implementation.

This issue tracks what's *not yet* covered.

### 1. Bbox-aware counts (design question)

Counts today are global. The legend says "OpenContext: 312,887" regardless of
viewport. In-map H3 summary tiles already show in-viewport counts. Should the
legend mirror that? (Lean: keep legend global — H3 tiles are the viewport-scoped
surface.)

### 2. Text-search-aware counts

When a search term is active, should facet counts reflect only matching samples?
The `facets_url` parquet has `label`, `description`, `place_name` columns — an
`ILIKE` predicate plugs into `buildCrossFilterWhere` cleanly. Decision needed
on whether to actually do it.

### 3. Hierarchical concept rollup

`context = earthinterior` is a sub-concept of `anysampledfeature`. Does the cube
roll children up to parents, or are counts strictly leaf-level? Need to inspect
the cube generation pipeline and verify against the live query path's behavior.
The two paths could disagree — that's a silent correctness bug.

### 4. Non-canonical axes

Cube schema covers only source / material / context / object_type. If the
explorer adds project / site / curation-location facets (issue #161 territory),
the cube needs schema migration *or* those axes fall back to live query only.
Worth deciding the contract before the next axis lands.

### 5. UX polish (lower priority)

- Threshold for showing "…" instead of just italic on slow queries
- Mobile collapse of long facet lists with counts
- "Reset" affordance when a user has heavily filtered

## Out of scope

- Refactoring `explorer.qmd` (#163 territory)
- Cube generation pipeline (separate repo)

## Acceptance

- Decisions recorded on items 1–4
- Items chosen for implementation broken into sub-issues

Cross-refs: #163, #164, #226. Supersedes #229.

[...]

PR #237, opened 2026-05-25
explorer: B1 viewport-aware facet counts (#234 step 3)

rdhyee:
## Summary

The facet legend now scopes per-source / per-material / per-context / per-object_type counts to the current map viewport bbox instead of the whole dataset. Pan or zoom → counts recompute. This resolves the silent disagreement between the legend ("global pivot tool") and the rest of the page ("Samples in View", table count, map dots) that the 2026-05-22 design session called out and that #234 picks as direction **B1**.

This is **step 3 of the #234 roadmap**, after #235 (facetNote URL-load, merged) and #236 (search real-count, merged). Subsequent steps (A1 search-as-global-filter, C3 auto-promote to point mode) are independent and build on this.

## Decisions locked with @rdhyee before implementation

| Q | Decision |
|---|---|
| Q1 — where do coordinates live for the bbox JOIN? | No existing parquet carries both the four facet columns and lat/lon. **JOIN `facets_url` to `lite_url` on `pid`** for this PR; follow-up to bake a pre-joined parquet if real-world latency proves bad. |
| Q2 — progressive refinement now? | **Single-pass full query**, debounced 250 ms via the existing `facetCountsReqId` infrastructure. Defer coarse-then-fine scaffold until latency shows it's needed. |
| Q3 — error-path semantics? | **Fall back to global baseline** via existing `applyFacetCounts(key, null)` — matches current no-filter behavior; user can re-pan to retry. |

## Architectural changes

- **New `isGlobalView()` helper** (altitude shortcut + rect-saturation check) decides when the cube fast-path stays valid. At alt > 10,000 km the camera is treated as global regardless of `computeViewRectangle` per-angle quirks, so legend totals don't jiggle on tiny camera rotations near the default zoom-out.
- **`updateCrossFilteredCounts`** snapshots `bboxSQL` once at function entry. Cube fast-path gated additionally on `bboxSQL === null`. The no-filter + non-global case now falls through to the slow path (B1's whole point: counts reflect what's in view even with zero facet filters).
- **Slow path** (when `bboxSQL` set): `JOIN facets_url f ↔ lite_url l ON l.pid = f.pid` and apply the bbox predicate against `(l.latitude, l.longitude)`. `buildCrossFilterWhere` takes an optional column-prefix arg so the JOIN-shape WHERE qualifies columns with `f.` to avoid ambiguity with `lite_url.source`.
- **`viewer.camera.moveStart`** → synchronous `markFacetCountsRecomputing()`. Italic-stale legend appears the instant the user starts panning, cleared when `applyFacetCounts` writes the post-moveEnd result.
- **`viewer.camera.moveEnd`** → `refreshFacetCounts()`. Reuses the existing 250 ms debounce + `facetCountsReqId` stale guard — bursts coalesce to one query, superseded in-flight queries discard their result on `await` resume.

## Test plan

- [x] `quarto render explorer.qmd` — passes
- [x] New `tests/playwright/facet-viewport.spec.js` — 3 specs pass locally
  - zoom-in → counts shrink relative to global, per-source counts ≤ global; zoom-out → restore to within 1 %
  - moveStart marks `.recomputing` synchronously (verified via in-page listener that runs after the explorer's)
  - negative control: cold global-view boot leaves no `.recomputing` stuck
- [x] Regression suite — `facetnote-url-load.spec.js` (2), `search-real-count.spec.js` (3), `url-roundtrip.spec.js` (5) — all pass (10/10)
- [x] `git diff --check upstream/main...HEAD` — clean
- [ ] Pre-deploy smoke gate on CI

## Known follow-ups

- **Latency not formally benchmarked.** The JOIN turns 4 parallel `GROUP BY` queries into 4 parallel `JOIN+GROUP BY` queries against the ≈300 MB `lite_url` parquet. If UX feels slow in real use, follow-up to bake a pre-joined facets-with-coords parquet and route the bbox path through it (Q1 fallback).
- **Cancellation-race spec deferred.** The `facetCountsReqId` stale guard is in place but not exercised by a Playwright test (two pans back-to-back inside the 250 ms window). Add if Codex flags it as a coverage gap.
- **Progressive refinement (coarse-then-fine)** explicitly out of scope — Q2 locked single-pass. File as a separate issue if latency benchmarks demand it.

## Cross-refs

- Closes the B1 row in #234
- Supersedes the bbox-scope portion of #230
- Unblocks step 4 (A1 search-as-global-filter)

---

Roadmap context: #234. Direction is **A1 + B1 + (C3-when-feasible, C2-with-warning)** with progressive refinement underlying the dynamic surfaces.

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -993,6 +993,50 @@ function viewerBboxSQL(latCol, lngCol, padFactor) {
     return ` AND ${latCol} BETWEEN ${south} AND ${north} AND ${lngClause}`;
 }
 
+// True when the current camera shows ≈the whole world (initial zoom-out
+// state), or when the camera can't compute a rectangle at all (off-globe).
+// In either case a viewport bbox predicate would be a no-op, so fast-paths
+// that depend on "no spatial constraint" (e.g. the pre-aggregated facet
+// cross-filter cube) remain valid and can be used.
+//
+// Used by `updateCrossFilteredCounts` (B1, issue #234 step 3) to decide
+// whether to gate the cube fast-path off and JOIN to `lite_url` for a
+// bbox-scoped slow-path query.
+//
+// Implementation note — why an altitude shortcut:
+//   At high altitudes Cesium's `computeViewRectangle` saturates at the
+//   sphere's extent for some camera angles but returns a tight bounding
+//   rect for others (e.g. at alt=5000 km over the equator the rect is
+//   ≈hemispheric, not full). The user's intuition is "max zoom-out = global
+//   counts," and at the explorer's default `globalRect` (which fits the
+//   `[-180,-60,180,80]` data extent and sits roughly at altitude ≈12000 km)
+//   the per-source legend totals should match the baselines, not jiggle on
+//   tiny camera rotations. The altitude check captures that "I'm zoomed all
+//   the way out" intent regardless of the exact rect Cesium reports. Below
+//   the altitude threshold we still let the rect vote "global" so flyTo
+//   to a true world-rect destination keeps working.
+//
+//   Constants live inside the function (not top-level `const`) because
+//   Quarto `{ojs}` cells reject top-level `const`/`let`/`var` and a bad
+//   declaration cascades — breaking every dependent cell (memory note
+//   `feedback_qmd_ojs_top_level`).
+function isGlobalView() {
+    const ALT_THRESHOLD_M = 1.0e7;  // 10,000 km
+    const LNG_PAD_DEG = 2;
+    const LAT_PAD_DEG = 2;
+    if (typeof viewer === 'undefined') return true;
+    let h = NaN;
+    try { h = viewer.camera.positionCartographic.height; } catch { /* mid-init */ }
+    if (Number.isFinite(h) && h > ALT_THRESHOLD_M) return true;
+    const b = paddedViewportBounds(0);
+    if (!b) return true;              // off-globe → no meaningful bbox
+    if (b.west > b.east) return false; // antimeridian-wrapping is never "global"
+    return b.west  <= -180 + LNG_PAD_DEG
+        && b.east  >=  180 - LNG_PAD_DEG
+        && b.south <=  -90 + LAT_PAD_DEG
+        && b.north >=   90 - LAT_PAD_DEG;
+}
+
 // === Cross-filter facet count UI helpers ===
 function applyFacetCounts(facetKey, countsMap) {
     const baseline = (viewer && viewer._baselineCounts) ? viewer._baselineCounts[facetKey] : null;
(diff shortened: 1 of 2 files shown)


rdhyee:
## Codex review iteration log (2 rounds → LGTM)

| Round | Codex finding | Resolution | Commit |
|---|---|---|---|
| 1 | (BLOCKING) `moveStart` only marked `.recomputing`; did not bump `facetCountsReqId` or clear the pending debounce. An older `updateCrossFilteredCounts` could resume mid-pan, pass the stale guard, write pre-move counts, and clear `.recomputing` — leaving the legend looking authoritative but stale until moveEnd's debounce fired the new query. | moveStart now `clearTimeout(facetCountsDebounce); ++facetCountsReqId;` in addition to marking the italic state. moveEnd's `refreshFacetCounts()` bumps the id AGAIN — supersession of already-superseded work, debounce coalesced. | `63b0457` |
| 1 | Test coverage gap: the spec exercised the no-filter bbox path but NOT the filter-active JOIN path — exactly the new SQL shape this PR introduces. | Added `bbox-aware count under an active source filter (JOIN + filter)`: activate a single source filter (cube fast-path at global view), capture filtered material counts, fly to Cyprus → forces slow path with `f.source IN (…)` JOIN. Assertion `cyprusTotal < filteredTotal` catches the silent fallback-to-baseline case (column-ambiguity SQL errors fall back to unfiltered globals, which would GROW back instead of shrinking — caught). | `63b0457` |
| 1 | 1 % restore tolerance was too loose. At `alt=15e6` the altitude shortcut forces the no-bbox baseline path, so restored counts should be value-by-value EXACTLY equal to the first global capture. | `expect(restoredCounts).toEqual(globalCounts)` — strict equality. | `63b0457` |
| 1 | Polish: moveStart comment missed the rationale for the id bump. | Comment expanded with the race-condition explanation. | `63b0457` |
| 1 | Documentation polish — high-altitude shortcut can return "global" for a non-fully-data-covering camera angle. | Intentional per the design rationale; no change. (Existing comment in `isGlobalView()` already explains the UX call.) | n/a |

**Round 2: LGTM** (Codex)

> Findings: none. I don't see a new correctness issue in `63b0457`. […] `moveStart` double-bump: safe. […] JOIN + filter test exercises the intended SQL shape because the checkbox state is changed synchronously before the Cyprus fly. […] Handler mechanics: `clearTimeout(null)` is safe, `facetCountsReqId` is initialized, and the mark/clear/bump ordering is fine under JS's single-threaded event execution.

Final test count: **14 passed** locally (4 new B1 specs + 10 regression: 2 facetnote-url-load + 3 search-real-count + 5 url-roundtrip).

**Known follow-ups deferred per Q1/Q2 plan decisions, not raised by Codex:**
- Latency was not formally benchmarked; baking a pre-joined facets+coords parquet (Q1 fallback) waits on real-world UX feedback.
- Cancellation-race Playwright spec (two pans back-to-back inside the 250 ms window) — not added; the `facetCountsReqId` stale guard is now exercised structurally by moveStart's invalidation, which Codex round 2 walked through and accepted.

Thanks codex for the round-1 catch on moveStart — the stale-and-not-italic interval was a real bug, not a theoretical one.

Issue #249, opened 2026-05-29
Architecture: should we refactor explorer.qmd before the next big feature? (analysis + options)


rdhyee:
## The question

With several substantial features recently landed or in flight in `explorer.qmd`
— heatmap mode (#241, #242, #246), the collection facet (#244), cross-filter
facets (#156), URL-state contract (#163/#164) — is the explorer codebase messy
enough that we should **refactor before adding the next big feature**, or keep
shipping incrementally?

This issue is the analysis to decide that, not a proposal to rewrite.

## Diagnosis (measured)

`explorer.qmd` is **~4,580 lines across 9 OJS cells**. The complexity is *not*
uniform — it's concentrated in one cell:

| Block | ~Lines | Health |
|---|---:|---|
| ~40 helper functions (`facetFilterSQL`, `readHash`/`buildHash`, card renderers, bbox math, …) | 730 | ✅ small, single-purpose |
| `viewer` cell (Cesium setup + click/hover handlers) | 270 | ✅ cohesive |
| `facetFilters` cell | 105 | ✅ fine |
| `tableView` cell | 455 | ⚠️ large but cohesive |
| **`zoomWatcher` cell** | **~2,200** | ❌ god-closure |
| `perfPanel` cell | 95 | ✅ fine |

**~48% of the file is a single closure.** `zoomWatcher` co-mingles: H3 cluster
loading (`loadRes`), point-mode sample loading (`loadViewportSamples`), mode
transitions (`enterPointMode`/`exitPointMode`/`applyLayerVisibility`), the
heatmap subsystem (`refreshHeatmap`/`renderHeatmap`/`clearHeatmap` + imagery
layer), camera (`moveEnd`/`camera.changed`) handlers, facet-count refresh, and
several stale-request guards. Recent features (heatmap, collection facet) mostly
landed in or adjacent to this cell.

### Why it's monolithic (it's the platform, not just neglect)

Quarto OJS has no ergonomic way to share **mutable** state across cells, so
stateful, tightly-coupled logic tends to collapse into one closure (the shared
`loading`, `requestId`, `loadResGen`, `heatmapReqId`, `currentRes`, etc.).
`EXPLORER_STATE.md` already documents the escape hatch — hoisting state onto
`viewer._globeState` / `viewer._heatmap*` so logic can be split across cells.

## The risk / tension

`explorer.qmd` is **dense with hard-won, issue-specific fixes**. A non-exhaustive
[...]

Issue #248, opened 2026-05-29
explorer: support searches of material samples described by a URI or PID identified concept


ekansa:
It would be very handy for the explorer to select a set of material sample records that are described by concepts identified by a URI or PID. 

For example, the URI `https://vocab.getty.edu/aat/300263796`  identifies a concept with the label  “whorls (spindle flywheels)”. It would be good to get a selection material sample records described by keywords that include concept. 

Another example would be the vocabulary concept `https://w3id.org/isample/vocabulary/materialsampleobjecttype/organismpart` (organism part). It would be good to get a selection of material samples records that have a sample object type for organism part, identified by that URI.

The  Interactive Explorer should have a new URL parameter to be used for such queries (where the concept URI or PID is the value passed in the URL parameter). Perhaps `described-by=` can be used as a URL argument for the Interactive Explorer?

rdhyee:
A1 (search-as-global-filter) just merged as #251 and is live on isamples.org, which gives this a clean foundation. A committed filter materializes a `search_pids` pid-set once, and every surface (samples table, globe points, facet counts, stats) constrains to it via a cheap semi-join. So #248 is essentially a **new producer** of that same pid-set.

Your two examples map to two flavors:

1. **iSamples-vocabulary URIs** (your organism-part example, `https://w3id.org/isample/vocabulary/materialsampleobjecttype/organismpart`) — these are exactly the URI-valued facet columns the explorer already carries (`object_type` / `material` / `context`). `described-by=<uri>` would build `search_pids` from an exact-URI match across those columns, and A1 then filters every surface automatically, with the concept's label resolved via the bundled `vocab_labels`. **Starting on this one now.**

2. **Arbitrary external concept URIs** (your Getty AAT "whorls" example, `https://vocab.getty.edu/aat/300263796`) — the sample records don't carry Getty URIs directly, so this likely resolves the concept URI → its label(s) and runs the free-text path on the label. That needs a URI→label resolution step beyond the bundled iSamples `vocab_labels`, so I'd propose treating it as a follow-on.

Will use `described-by=` as the URL parameter as you suggested, starting with flavor 1.

Issue #247, opened 2026-05-29
explorer: free-text search does not filter the main samples table — "N match the current filters" shows unrelated samples (A1 gap)


rdhyee:
## Summary

A free-text search (e.g. `bucchero`) leaves the **main "Samples table" and its "N samples match the current filters" count showing samples that have nothing to do with the search term.** The search box flies the camera to the right place and the side-panel list shows the correct hits, but the most prominent results surface — the table under the map — fills with whatever happens to be in the new viewport. The label "match the current filters" is actively misleading, because **the search term is not one of the filters.**

This is a concrete, user-facing instance of the **A2 → A1 gap** already captured architecturally in #234 (roadmap **Step 4 / A1**: "search as global filter"). Filing as a standalone reproducible bug so there's something crisp to point at and verify a fix against.

## Reproduction

URL (production):

```
https://isamples.org/explorer.html?search=bucchero&sources=OPENCONTEXT%2CGEOME%2CSMITHSONIAN#v=1&lat=40.8958&lng=5.2905&alt=3048966
```

(`bucchero` = Etruscan pottery; the canonical site is Poggio Civitate / Murlo, Tuscany.)

Observed once loaded:

| Surface | Result | Correct? |
|---|---|---|
| Camera | Flies to Poggio Civitate / Murlo, Tuscany (≈43.15, 11.40) | ✅ |
| Side-panel list (`samplesSection`, right column, **below the fold**) | Top-50 = OpenContext "Pottery PC…/Object PC…" — genuine bucchero finds, all at (43.15, 11.40) | ✅ but easy to miss |
| Search-results line | `50 of 2,693 results for "bucchero"` | ✅ |
| **Main "Samples table"** (the big one under the map) | **"43,803 samples match the current filters"**, every visible row a **GEOME marine mollusk** (Turritella, Cerithium, Bittium) in the Tyrrhenian Sea. **0 rows are bucchero.** | ❌ |

So a user who types `bucchero` and looks at the obvious results table sees 43,803 sea snails described as matching their filters.

## Root cause

The main table + its count are driven by `loadCount()` / `loadPage()` (`explorer.qmd`, ~L2091–2155). Their WHERE clause is **only**:

```sql
WHERE pid IS NOT NULL
  ${sourceFilterSQL('source')}   -- source checkboxes
  ${facetFilterSQL()}            -- material / context / object_type
  ${bboxSQL}                     -- viewport
```

[...]

PR #251, opened 2026-05-31
explorer: search as a real global filter (#234 Step 4)

rdhyee:
## A1: search as a real global filter (#234 Step 4)

Makes a committed free-text search a **global filter** across every explorer surface, instead of an optional side-panel lookup. On search, `buildSearchFilter()` materializes a non-temp DuckDB table `search_pids` (one ILIKE scan over `sample_facets_v2`), and every surface then constrains via a cheap semi-join `AND pid IN (SELECT pid FROM search_pids)`:

- **Samples table** filters to the search ("N of M in this map view").
- **Globe** enters point mode and renders only the matching dots (clusters can't be text-filtered).
- **Facet legend counts** and **stats** scope to `search ∩ viewport`.

Fixes the incoherence reported in **#247** (table claimed unrelated viewport samples "match the current filters" during a search; interim honesty fix shipped as #250, this completes it).

### What's verified
- **Globe coherence** — `bucchero` → 2,693 pids, point mode, h3 cleared, `samplePointsLen ≤ total`. Confirmed headed + headless **and on a live deploy** (real HTTP/2 + 206 ranges, prod data).
- **Search perf** — collapsed an A1 double-scan (pid-set build + side-panel results both scanned the 63 MB facets); the side panel now reads the materialized `search_pids`, so it's **one facets scan per search**, matching pre-A1. CI smoke gate (`pottery`) passes.
- **Filter coherence** — facet legend counts now use the same padded viewport (`VIEWPORT_PAD_FACTOR`) as the table/heatmap/point-loader/stat, so legend == "N match" (was reading low: ~166 vs ~481 at a wide view).
- **Production-clean** — all A1 debug instrumentation (`a1dbg`/`__a1log`/`__a1globe` + on-page panel) is gated behind `?debug=a1`; default load has a clean global namespace and no overhead.

### Included commits
data_base dev-override fix · debug gating + dev-probe removal · double-scan collapse · facet-padding coherence fix · dev verify infra (`dev_server.py`, `tests/playwright/a1-verify.mjs` + probes). The dev/test infra is optional — happy to drop `dev_server.py` / the probes if you'd prefer them out of upstream.

### Known / deferred (not blockers, flagged for review)
- **Heatmap** isn't search-aware yet (`renderHeatmap` omits `searchFilterSQL`), so the "filtered density" layer stays unfiltered under a committed search. Follow-up.
- **Selection revalidation** on search change (clear a selection that's no longer in the filtered set).
- **Cold-search latency** — A1 moves the un-indexed full-text ILIKE scan to the front of the common flow; the proper fix is the BM25 substrate (#168–172). The "Building search filter…" affordance masks it for now.

### Relates to
#234 (umbrella), #247/#250, #248 (concept-URI search — a second producer of the same `search_pids` set), #249 (the "refactor explorer.qmd first?" question — this PR is a data point for it).

### Staging
Deployed and verified on the rdhyee fork's GitHub Pages (same data/infra as isamples.org). Suggest **squash-merge** — the branch carries some WIP commits whose messages predate the fixes.

### DIFF
--- a/A1_SCOPING.md
+++ b/A1_SCOPING.md
@@ -1,0 +1,129 @@
+# A1 Scoping — "search as a global filter" (#234 Step 4)
+
+Goal: when a free-text search is committed, **every count/where surface reflects `searchTerm ∩ viewport ∩ source/facet filters`**, not just the side-panel results list. Currently search only drives a side list + camera fly; the table, point loader, facet counts, stats, and globe ignore it (axis A2). #250 (interim) only relabels the table; this is the real fix.
+
+## 1. Data-model reality (the constraint that shapes the design)
+
+The search predicate (`textSearchWhere`) matches **3 columns**: `label`, `description`, `CAST(place_name AS VARCHAR)` — all read from **`facets_url`** (`sample_facets_v2.parquet`). `description` was deliberately moved to facets_url in #168 *because* the search needed it.
+
+| Surface | Function | Reads from | Has search cols? |
+|---|---|---|---|
+| Samples table (count + page) | `loadCount` / `loadPage` | `lite_url` (samples_map_lite) | **label + place_name only — NO `description`** |
+| Point-mode dots | `loadViewportSamples` | `lite_url` | NO description |
+| Facet legend counts | `updateCrossFilteredCounts` | `facets_url` (global) **or** `facets_url ⋈ lite_url` (bbox path) | **YES — facets_url has all 3** |
+| Facet counts cube fast-path | (cross_filter parquet) | `cross_filter_url` (pre-aggregated) | **NO — globally pre-aggregated, cannot be text-filtered** |
+| Globe clusters | H3 summary parquets | `*_h3_summary_res{4,6,8}` | **NO — pre-aggregated, `dominant_source` only** |
+| Heatmap | wide/lite | — | NO description |
+| "Samples in View" stat | derived from above | — | follows its surface |
+
+Two hard truths:
+1. **`description` is only in facets_url.** Any surface that queries `lite_url` (table, points) must reach facets_url to get full search parity — i.e. a JOIN — or accept reduced recall (label+place_name only).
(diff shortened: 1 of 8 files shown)


rdhyee:
## Multi-AI review cycle → dual approval

Ran an iterative Codex (gpt-5.4) review/revise loop on this PR alongside Claude's own review. Codex found genuine issues across three rounds; all fixed and empirically re-verified (the `a1-verify.mjs` coherence harness stayed green throughout, and overlap/clear cycles were checked directly):

| Round | Finding | Fix | Commit |
|---|---|---|---|
| 1 | **Staging-table race** — a fixed `search_pids_next` name let two overlapping searches cross-contaminate (a later build could swap an earlier build's pids into `search_pids` under its own term; token checks guarded the publish, not the shared staging object) | token-scoped `search_pids_next_${token}`, dropped in `finally` | `a576dea` |
| 1 | **Heatmap search-blind** — `renderHeatmap` omitted `searchFilterSQL` and `heatmapFilterHash` omitted the search token, so the density overlay stayed unfiltered under a committed search | append the semi-join + add the active-search token to the hash | `a576dea` |
| 2 | **Stale direct-reader** — leaving `search_pids` in place on clear let the side-panel `SELECT FROM search_pids` render the *previous* term's rows on a build failure | replace with an empty same-shape table on clear (chosen over an early-return guard so the `#167` telemetry `finally` still runs) | `8a9a1d3` |
| 3 (nit) | a genuine build failure read as "No results" | `searchFilterBuildFailed` flag → "Search error" | `0a91361` |

Verified:
- `a1-verify.mjs`: ✅ A1 COHERENT (headed + headless + live rdhyee deploy).
- Overlap: `bucchero`→`soil` back-to-back publishes `soil`/2969 (its own count), not bucchero's under soil's term.
- Clear cycle: `bucchero`→clear→`soil` clean, no stale rows.

**Verdicts:** Claude — approve; Codex — **APPROVE** (no remaining requested changes). Deployed and verified on the rdhyee fork's GitHub Pages (same data/infra as isamples.org).

Issue #255, opened 2026-06-01
explorer: #248 described-by — add a removable concept chip + hide stale 'Click a cluster' placeholder


rdhyee:
## Why
`described-by=<uri>` has no UI surface (URL-only) and its results render below the fold. Two small fixes:
1. A pinned **'Described by: <label> ✕' chip** at the top of the side panel — visible concept identity + one-click clear. (This is the UI element users will look for; also helps because the concept filter is currently fragile — any search-box entry silently replaces it.)
2. **Hide the '#clusterSection "Click a cluster or sample" placeholder'** while a search/concept is active — it's meaningless then and eats the vertical space the results want.

## Scope
- `explorer.qmd`: `doDescribedBy()` already sets `window.__searchFilter = {kind:'concept', uri, term:<label>, ...}` — render the chip from that; clear = call the existing clear path + remove `described-by` from the URL.
- The chip should also show for a future paste-a-URI input (Flavor B). Escape the label with `escapeHtml` (see XSS fix in #252).
- Hide `#clusterSection` when `searchIsActive()`.

## Acceptance
- With a concept active, a chip shows the label + ✕ that clears it; placeholder hidden during search; Codex-reviewed; staging-verified.

rdhyee:
Implemented for #255 and pushed as a draft PR on the origin fork: https://github.com/rdhyee/isamplesorg.github.io/pull/6

What changed:
- Added a pinned side-panel chip for active `described-by` concept filters: `Described by: <label> ✕`.
- Rendered the chip from `window.__searchFilter` when `kind === 'concept' && active`, escaping the label with the existing `escapeHtml()` helper before `innerHTML`.
- Wired the chip clear button through the existing clear path: `clearSearchFilter()` + `applySearchFilterChange()` + URL rewrite via `writeQueryState({ commitText: '' })`.
- Hid `#clusterSection` whenever `searchIsActive()` is true, and restored it after clearing.
- Kept `doSearch()`'s hot path untouched.

Validation:
- `quarto render explorer.qmd`
- Served `docs/` on port 8079.
- Headlessly loaded `?described-by=https://w3id.org/isample/vocabulary/materialsampleobjecttype/1.0/wholeorganism` and asserted the chip showed `Whole organism material sample` plus `✕`, the cluster placeholder was hidden, and clicking `✕` cleared both the filter and the URL param.
- `git diff --check`

[...]

Issue #254, opened 2026-06-01
explorer: make the map stats panel (Resolution / Clusters / Samples / Load time) collapsible, collapsed by default


rdhyee:
## Why
The 4-box stats panel (Resolution, Clusters Loaded, Samples Loaded, Load Time) is great for power users but confusing for novices (RY, 2026-06-01). Make it **collapsible, collapsed by default**, with a clear expand affordance.

## Scope (self-contained, low risk — good parallel task)
- The `.stats-compact` grid lives in the first `.panel-section` of `.side-panel` in `explorer.qmd` (~L600s). Element ids: `#sPhase`, `#sPoints`, `#sSamples`, `#sTime`.
- Add a small header/toggle (e.g. a 'Stats ▾' disclosure) that shows/hides the four boxes. **Default collapsed.**
- Optional: persist open/closed in localStorage and/or a URL param, mirroring the facet sections' pattern (the Material/Sampled-Feature/Specimen-Type `.filter-header` onclick toggle is a copyable model).
- Keep updating the values while collapsed (cheap) so expanding shows current numbers.

## Acceptance
- On load, stats are collapsed; clicking the disclosure reveals them; state is obvious. No layout shift that buries the results panel further. Smoke gate green.

rdhyee:
Shipped to isamples.org via #261 (stats panel collapsed by default).

PR #258, opened 2026-06-01
Fix facet count race after search flyTo

### LINKED ISSUE #253: explorer: facet-legend counts go stale after a committed search (recompute race, not a missing filter)
## Summary
After committing a free-text (or concept) search, the **facet-legend counts** — most visibly the always-on **source counts** — keep showing **viewport-only** numbers that ignore the active search. They are **not** a brief transient: observed stuck for **≥7 s** (with `explorer-busy` already cleared). They correct to the proper search-filtered values only after the user **interacts with a facet section**, which forces a fresh recompute.

This is the visible residue of the "bucchero pulls SESAR" report (2026-05-29 tech call). The #250 fix corrected the *results/table* ("N of N matches in this map view" is right; result rows are OpenContext-only). The **legend counts** were not fully covered.

## Reproduce (staging or isamples.org)
1. `…/explorer.html#v=1&lat=42.5&lng=12.0&alt=400000`
2. Search `bucchero` (world scope, the default — Enter in the search box).
3. Watch the **source legend**: it shows e.g. **`SESAR (2,573)`** and stays there.
   - Ground truth: bucchero = 2,693 samples, **100% OpenContext, 0 SESAR**. So the legend should read `SESAR (0)`, `OpenContext (~2,670 in view)`.
4. Now expand any facet section (Material / Sampled Feature / Specimen Type) → a recompute fires → source corrects to **`SESAR (0)`** and the expanded dimension reads the bucchero-filtered breakdown (e.g. object_type `artifact (~2,289)`, `othersolidobject (0)`; material `anthropogenicmetal (~2,660)`, `earthmaterial (0)`).

## Root cause (important — the SQL is already correct)
`updateCrossFilteredCounts()` (explorer.qmd ~L2887) **does** apply the `search_pids` semi-join via `searchFilterSQL('f.pid')` in its slow path (~L2994), and correctly gates off the cube/baseline fast-paths when `searchIsActive()` (L2930, L2945). `source` is one of its `dims` (L2855). So when this function runs with the search active, the counts are right.

The bug is **sequencing**: a committed world-scope search auto-flies the camera to the first result, and that flight's `camera.changed` / `moveEnd` cascade appears to leave the legend showing a pre-search / in-flight viewport recompute rather than a final search-aware one. The committed search does not reliably trigger a last, un-superseded `refreshFacetCounts()` after the flyto settles.

## Suggested fix direction
Ensure a committed search fires a final, un-superseded search-aware count recompute *after* the auto-flyto settles (and/or have the flyto-triggered recompute read the now-active search state). Verify across world + area scope and the `described-by=` concept producer (#248), which shares the same path.

## Decisive test not yet run
An **area-scope** search (no auto-flyto): if the legend settles to `(0)` correctly with no camera flight, that confirms the flyto as the culprit.

Context: A1 #234 step 4 / #251; cross-source results fix #250; concept filter #248 (#252).

rdhyee:
## Summary
- Add a one-shot post-flyTo facet-count refresh for committed world-scope searches so the final request supersedes camera moveEnd recomputes.
- Preserve existing A1 search-aware SQL and B1 viewport-aware count behavior.
- Add a Playwright regression covering the area-scope no-fly diagnostic and the world-scope flyTo path without expanding facets.

## Verification
- `/opt/homebrew/bin/duckdb` confirmed `bucchero` ground truth: 2,693 total, 2,693 OpenContext, 0 SESAR, 0 earthmaterial, 0 othersolidobject.
- `/usr/local/bin/quarto render explorer.qmd`
- `HEADLESS=1 A1_CLOSE=1 A1_BASE='http://localhost:8077/explorer.html?data_base=/data&debug=a1&sources=OPENCONTEXT%2CGEOME%2CSMITHSONIAN' node tests/playwright/a1-verify.mjs`
- `TEST_URL=http://localhost:8077 ./node_modules/.bin/playwright test tests/playwright/facet-search-count-race.spec.js --project=chromium --reporter=line`

Fixes #253

### DIFF
--- a/A1_SCOPING.md
+++ b/A1_SCOPING.md
@@ -1,0 +1,129 @@
+# A1 Scoping — "search as a global filter" (#234 Step 4)
+
+Goal: when a free-text search is committed, **every count/where surface reflects `searchTerm ∩ viewport ∩ source/facet filters`**, not just the side-panel results list. Currently search only drives a side list + camera fly; the table, point loader, facet counts, stats, and globe ignore it (axis A2). #250 (interim) only relabels the table; this is the real fix.
+
+## 1. Data-model reality (the constraint that shapes the design)
+
+The search predicate (`textSearchWhere`) matches **3 columns**: `label`, `description`, `CAST(place_name AS VARCHAR)` — all read from **`facets_url`** (`sample_facets_v2.parquet`). `description` was deliberately moved to facets_url in #168 *because* the search needed it.
+
+| Surface | Function | Reads from | Has search cols? |
+|---|---|---|---|
+| Samples table (count + page) | `loadCount` / `loadPage` | `lite_url` (samples_map_lite) | **label + place_name only — NO `description`** |
+| Point-mode dots | `loadViewportSamples` | `lite_url` | NO description |
+| Facet legend counts | `updateCrossFilteredCounts` | `facets_url` (global) **or** `facets_url ⋈ lite_url` (bbox path) | **YES — facets_url has all 3** |
+| Facet counts cube fast-path | (cross_filter parquet) | `cross_filter_url` (pre-aggregated) | **NO — globally pre-aggregated, cannot be text-filtered** |
+| Globe clusters | H3 summary parquets | `*_h3_summary_res{4,6,8}` | **NO — pre-aggregated, `dominant_source` only** |
+| Heatmap | wide/lite | — | NO description |
+| "Samples in View" stat | derived from above | — | follows its surface |
+
+Two hard truths:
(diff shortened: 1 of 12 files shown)


rdhyee:
Superseded by #261, which shipped the consolidated runtime (stats-collapse #254 + chip #255 + facet-count fix #253) to isamples.org. Closing the PR. NOTE: the regression tests from these branches are not yet on upstream — they'll be re-added via one clean follow-up PR (incl. repackaging the count-race spec as a standalone .mjs to fix the [P2] default-suite/`/data` dependency).

PR #259, opened 2026-06-01
[codex] Add described-by filter chip

### LINKED ISSUE #255: explorer: #248 described-by — add a removable concept chip + hide stale 'Click a cluster' placeholder

rdhyee:
## Summary
- Add a pinned side-panel chip for active `described-by` concept filters, rendered from `window.__searchFilter` with the label escaped via `escapeHtml()`.
- Wire the chip clear button through the existing search-filter clear path and remove `described-by` from the URL.
- Hide the cluster placeholder while any search/concept filter is active and restore it after clearing.

## Validation
- Served `docs/` on port 8079 and headlessly loaded `?described-by=https://w3id.org/isample/vocabulary/materialsampleobjecttype/1.0/wholeorganism`; asserted the chip shows `Whole organism material sample` plus `✕`, `#clusterSection` is hidden, and clicking `✕` clears the filter plus URL parameter.
- `git diff --check`

Fixes #255

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -219,6 +219,45 @@ format:
         max-height: var(--explorer-map-height);
         overflow-y: auto;
     }
+    .filter-chip-host {
+        position: sticky;
+        top: 0;
+        z-index: 3;
+    }
+    .filter-chip-host[hidden] { display: none; }
+    .described-by-chip {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        gap: 8px;
+        background: #ede7f6;
+        border: 1px solid #d1c4e9;
+        border-radius: 6px;
+        color: #5e35b1;
+        font-size: 12px;
+        font-weight: 600;
+        padding: 8px 10px;
+    }
+    .described-by-chip-label {
+        min-width: 0;
+        overflow-wrap: anywhere;
+    }
+    .described-by-chip-clear {
+        flex: 0 0 auto;
+        background: transparent;
+        border: 0;
+        border-radius: 4px;
+        color: inherit;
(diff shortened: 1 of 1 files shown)


rdhyee:
Superseded by #261, which shipped the consolidated runtime (stats-collapse #254 + chip #255 + facet-count fix #253) to isamples.org. Closing the PR. NOTE: the regression tests from these branches are not yet on upstream — they'll be re-added via one clean follow-up PR (incl. repackaging the count-race spec as a standalone .mjs to fix the [P2] default-suite/`/data` dependency).

Issue #265, opened 2026-06-05
Issues with material types


akthom:
The list of material types in the explorer does not look accurate compared to what we had in the past version of iSamples Central. 

<img width="349" height="425" alt="Image" src="https://github.com/user-attachments/assets/f891d3ce-d55b-4322-8dd9-baf1733477d7" />

For example, there's a material type that's listed as "material". That doesn't make sense at all -- that's not a material. It seems like either a) the material type metadata in the geoparquet files is incorrect, or b) something is not being read correctly from the files.

akthom:
Related to issue https://github.com/isamplesorg/isamplesorg.github.io/issues/260

rdhyee:
You're right, Andrea — and your instinct (b), "something isn't being read correctly," is the closer one. Two things are going on:

1. **A read bug (the main one):** the build that produces the facet data picked the *first* entry of each sample's material list (`p__has_material_category[1]`). But source records — especially SESAR — carry the **whole SKOS ancestry** in that list, and the broad root concept (`.../material/1.0/material`, whose label is literally "Material") can sit in the first slot. So ~half a million samples got tagged with the meaningless root. That's exactly the "material" entry you saw.
2. **Some genuinely-unclassified samples:** a chunk of records are only ever classified at that top-level "material" concept upstream — those are real, just unclassified.

I verified against the 2026-01 wide data: **528,983** samples had the root in slot 1.

**Fix (PR #271):** resolve the full list, drop the root, and take the first *real* concept. It's deliberately conservative — only the affected samples change (~318k get a real material; ~210k that are root-*only* become unclassified and drop out of the facet), and **zero** samples keep "Material", so that bogus entry disappears entirely. Correctly-classified samples are untouched.

Two honest caveats:
- This picks the first *non-root* concept, not necessarily the most-*specific* one (the lists aren't clean hierarchy paths — some even end in an unrelated "rock"). Truly picking the leaf needs the SKOS hierarchy; I've noted that as a follow-up.
- It's a build-time change, so it shows up on the site only after the derived parquet is rebuilt and redeployed.

---
*— 🤖 rbotyee (RY's bot). RY skimmed this; the answer is the bot's. Ping @rdhyee if it's off.*

rdhyee:
## ✅ Fixed in production — closing with evidence

@akthom — your instinct was right on both counts, and both are now fixed on **isamples.org**:

[...]

PR #271, opened 2026-06-05
build: stop SKOS root 'Material' leaking into the material facet (#265)

rdhyee:
Addresses #265 (the bogus 'material' material-type).

**Cause:** `base_samples_sql` picked `p__has_material_category[1]`, but source arrays (esp. SESAR) carry the full SKOS ancestry and the broad root (`.../material/1.0/material`, label 'Material') can sit at position 1.

**Fix:** resolve the whole concept array, drop the root, take the FIRST remaining concept. Conservative — only samples whose `[1]` was the root change. Verified against the 2026-01 wide file:
- 528,983 root-`[1]` samples change: ~318k get a real concept, ~210k are root-only → NULL (excluded from the facet)
- 0 samples retain the root → the bogus 'Material' facet disappears
- correctly-classified samples untouched, incl. Eric's ceramic `ark:/28722/k2p55x96j` (stays `anthropogenicmetal`, **not** flipped to a deeper-but-wrong array entry like 'rock')

**Limits:** this does *not* pick the most-specific concept — the arrays aren't clean SKOS paths (some OC arrays end in an unrelated 'rock'), so true leaf selection needs the SKOS hierarchy (follow-up). `context`/`object_type` roots left as-is for now.

⚠️ **Stacks on #264** (which introduces `build_frontend_derived.py`) — the diff will collapse to one file once #264 merges. Also note this is a build-time change: it takes effect only when the derived parquet is rebuilt + redeployed.

---
*— 🤖 rbotyee (RY's bot). RY skimmed; the bot did the work. Ping @rdhyee.*

### DIFF
--- a/DATA_PROVENANCE.md
+++ b/DATA_PROVENANCE.md
@@ -1,0 +1,74 @@
+# iSamples Explorer — Data Provenance
+
+How every parquet file the explorer uses is generated, from root to publish.
+*Reviewed 2026-06-02 (CC, via codebase audit). Complements `SERIALIZATIONS.md` (format/schema reference); this file is the end-to-end build chain + the automation gaps.*
+
+> **Load-bearing constraint:** the **root export cannot be regenerated.** It was produced from the iSamples Central Solr API (`central.isample.xyz`), **offline since Aug 2025**. The Zenodo-archived export is a **frozen root**. Any *new* data (e.g. concept URIs, thumbnails) therefore must come from a **per-source supplementary file merged into the base by `pid`** — the "sidecar" pattern (see Stage 3) — not from re-exporting.
+
+## Pipeline DAG
+
+```
+Source collections (SESAR · OpenContext · GEOME · Smithsonian)
+   │  iSamples Central Solr API  ── OFFLINE since Aug 2025 (cannot re-run) ──┐
+   ▼                                                                         │
+STAGE 0/1  export_client → JSONL → GeoParquet                               │ frozen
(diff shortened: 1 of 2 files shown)


rdhyee:
⚠️ **Do not merge as-is** — this build change has a **performance regression** found by actually running it. Mixing the new `MAP` cross-join (material) with the original correlated subqueries (`context`/`object_type`) blows up the planner: a full rebuild from the 202604 wide sat in "building base sample tables" for **>16 min** before I killed it.

Also: the script assumes `geometry` is WKB BLOB (`ST_GeomFromWKB`) — it throws `BinderException` on `GEOMETRY`-typed wides (202601 / Zenodo).

The material-selection *logic* is correct (validated: root → 0, `k2p55x96j` preserved), but the builder needs hardening before it can produce files. Tracked in the new pipeline epic (https://github.com/isamplesorg/isamplesorg.github.io/issues/273), workstream 2. Static review (incl. Codex) did not catch either — only execution did.

— 🤖 rbotyee (RY directing)

rdhyee:
Superseded by #274 — the material first-non-root fix is folded into the hardened, tested builder there (the standalone change here had a perf regression + geometry contract bug that only surfaced on execution). Recommend closing this in favor of #274. — 🤖 rbotyee

rdhyee:
Superseded by #274 (merged): its first-non-root material selection logic was folded into `scripts/build_frontend_derived.py` with fixture tests + the semantic validation gate. Closing unmerged.

— 🤖 rbotyee (RY's bot)

Issue #306, opened 2026-06-19
Data: sample_facet_masks omits ~29,917 located samples with no tree membership (built from membership, not samp_geo)


rdhyee:
## Summary

`sample_facet_masks` (and the `sample_facet_membership` it derives from) **omit ~29,917 located samples** that have coordinates but no hierarchical facet membership. Any count or filter computed *from the masks* therefore silently undercounts the true located universe.

## Evidence

Measured on the deployed 202608 data:

- `facets_v3` (located pid universe): **6,026,242** pids
- `sample_facet_masks` / membership pid universe: **5,996,325** pids
- **Missing from masks: 29,917 located pids**

## Root cause

`build_sample_facet_masks()` ([scripts/build_frontend_derived.py:537](https://github.com/isamplesorg/isamplesorg.github.io/blob/main/scripts/build_frontend_derived.py)) starts from `membership`:

```sql
FROM membership m JOIN nb ON ...
GROUP BY m.pid
```

A located sample that asserts **no** concept in any of the three SKOS trees (material / context / object_type) never appears in `membership`, so it gets **no mask row at all** — rather than a row with all-zero masks. It's invisible to the bitmask path.

## Why it matters now

It's latent today (the masks are only used for *filtering*, where a no-membership sample legitimately matches no tree node). But the #305 plan computes **facet counts and the located-sample baseline** from a masks-based index. Those counts must reflect the full located universe (incl. zero-membership samples, e.g. the "source" facet count and any "samples in view" total), so this gap would produce convincing-but-wrong numbers.

## Fix

Build the per-pid index **starting from `samp_geo`** (the located universe), LEFT JOIN the aggregated masks, and emit **zero masks** for samples with no hierarchical membership:

```
samp_geo  LEFT JOIN (aggregated masks)  ->  COALESCE(mask, 0)
```

Carry this into the new complete index described in #305 Phase 1 (`pid, source, material_mask, context_mask, object_type_mask, build_id, schema_version`), and add a validator gate asserting `index.pid_count == facets_v3.pid_count`.

## Verification

- Validator: pid-set equality between the new index and `facets_v3`; source equality; mask≡membership for samples that DO have membership.
- Confirm the 29,917 previously-missing pids appear with all-zero masks.

Found during the #305 plan review (Codex). Related: #305, #299 (introduced the masks), #304.

rdhyee:
[...]

Issue #305, opened 2026-06-19
Meta: facet-count correctness & coherence (counts beyond the single-filter cube)


rdhyee:
## Why this issue exists

Facet-count correctness keeps surfacing as separate bug reports (#296, #304, and others from collaborators), but they share **one root cause** and **one fix lever**. This is the higher-level tracker so the team can see the whole picture in one place and not re-diagnose each report from scratch.

It effectively **succeeds #234** (now closed) for the *counts* dimension: #234's "coherent filter semantics across all surfaces" roadmap shipped, but the count-coherence frontier outlived it.

## The through-line (plain English)

The Explorer can show **correct, cross-filtered facet counts instantly for exactly one active filter** — that case is served by a precomputed lookup ("the cube", #290/#298). Everything *beyond* one filter currently falls off that fast path:

- **More than one filter** at the zoomed-out view → other facets revert to their unfiltered global totals (this is #304).
- **Viewport-scoped** counts, **active search**, and **flat mode** also fall through to live paths with their own rough edges.

So the mental model is: *the single-filter cube is the spine; the frontier is "counts when more than one thing is true at once."*

Crucially, the **map/point filtering is correct** in all these cases — it's specifically the **count numbers** that mislead. And **zooming in** makes counts correct (the live membership path cross-filters by all active dims). The breakage is specific to the global/zoomed-out view.

## The shared fix lever

We already shipped the tool to fix this: the **`sample_facet_masks` bitmask** (#299) computes multi-filter results in one ~6M-row columnar scan, instead of the ~39M-row `sample_facet_membership` scan that originally forced the global multi-filter count path to be stubbed to the baseline. The plan is to un-stub the global multi-filter count path on top of the masks.

## Children / related

- **#304** — facet counts wrong when filtering by multiple facet values (the live bug; keep riffing there). **← primary**
- **#230** — facet counts: remaining work after #157/#226 (the concrete leftovers list)
- **#276** — material facet semantics: single "first non-root" value vs membership ("anywhere in array") counts (a *semantics* decision that shapes what the counts even mean)
- **#286** — clearing a search doesn't deterministically restore viewport-scoped counts (needs repro)

## What "done" looks like

- Facet counts reflect **all** active selections (multi-filter), at any zoom, served fast (masks-backed) rather than reverting to baseline.
- A single, stated **semantics contract** for what a count means (resolves #276), applied consistently across map / table / facets.
- The individual children above closed against that contract.

[...]

PR #308, opened 2026-06-20
#304/#305 Phase 2: correct multi-filter facet counts at global view (stacked on #307)

### LINKED ISSUE #304: Interactive Explorer: Facet counts wrong when filtering by multiple facet values
The Interactive Explorer appears to work as expected when one filters by 1 facet value. The facet counts update appropriately.

However, when one applies multiple facet values as filters (for example when one filters by "Material: Anthropogenic material" and "Specimen Type: Artifact" then the facet values are wrong.

rdhyee:
> ## In plain English
>
> **The problem this fixes:** On the zoomed‑out world view, when you picked **two or more
> filters at once** (e.g. Material = *Anthropogenic* **and** Object type = *Artifact*), the
> little count next to each facet was **wrong** — it ignored your other selections and showed
> the unfiltered total instead. The map dots were always filtered correctly; only the **numbers**
> next to the facets misled.
>
> **What it does now:** Those counts correctly reflect **all** of your active filters together.
> Concrete example (Eric/John's case): *Anthropogenic + Artifact* now reads **145,770** (correct)
> instead of the old **261,086** (the unfiltered‑artifact total — the bug).
>
> **A safety rule built in (the "honesty rule"):** if the correct cross‑filtered count ever
> can't be computed (data file not yet published, or a query hiccup), the UI shows an explicit
> dash **"(—)"** rather than quietly falling back to a misleading number.
>
> **Why it's safe:** it adds a new, more‑complete data file (one row per located sample) and a
> careful set of readiness checks; if anything is off, it degrades to the honest dash. Reviewed
> independently by Claude and Codex, and verified against the real production data.
>
> ---
>
## Phase 2 of #305 — fix #304: correct multi-filter facet counts at global view

**🔗 Stacked on #307 (Phase 1).** Draft until #307 merges; the net-new change here is **`explorer.qmd` only** (Phase 1's build/validator commits show until the base merges).

Fixes **#304**: at the global/zoomed-out view, facet **counts** only cross-filter for a *single* active filter (served by the `facet_tree_cross_filter` cube, #298). With **2+ active selections**, `effectiveSingleFilter()` returns null, the cube is skipped, and `updateCrossFilteredCounts` falls to a legacy early-return that reverts every tree dim to the **unfiltered baseline** — misleading counts. (Map/point filtering was already correct; only the numbers misled.)

### The fix
A new path in `updateCrossFilteredCounts`, between the single-filter cube and the baseline early-return, routes the **global / full-tree-mode / no-search / multi-filter** case through the complete per-pid index (`sample_facet_index`, #305/#306) using the `facet_node_bits` bitmask predicate:

- **`directFilterSnapshot()`** reads all selections from the controls **without** zeroing tree dims at global view (that zeroing in `describeCrossFilters()` is the root cause).
- **`maskIndexWhere()`** builds the per-dim cross-filter predicate (OR within a dim via the bit mask, AND across dims, **exclude-self**); source-impossible zeroes tree targets but not source's own histogram.
- **`applyMaskIndexCounts()`** runs **one `UNION ALL`** statement (one DuckDB-WASM connection slot, not 4 concurrent scans), then applies all dims **atomically** after a single `facetCountsReqId` stale check. `COUNT(*)` over the one-row-per-pid index == distinct pids.

### Honesty rule (never baseline under active filters)
On any failure, or while the index is unpublished/inconsistent, multi-filter counts show an explicit **"(—)"** dash (`.count-unavailable`), **never** the misleading baseline. A `facetIndexReady` preflight gates on: index present + `schema_version` + `build_id` membership-half == `node_bits` generation + **per-source coverage** == the located universe. When readiness settles after boot, `__onFacetIndexReady` reconciles any active filter.

### Verified on real 202608 data
- Built + validated the real `sample_facet_index` (validator **ALL-PASS**); it covers **6,026,242** located pids vs masks' 5,996,325 — exactly the **29,917** #306 samples recovered.
- **Eric's exact case** (Material: *anyanthropogenicmaterial* + Object type: *artifact*): the mask-index path returns the correct cross-filtered **145,770** vs the buggy baseline **261,086**.
- Native query time 3–8ms; the single `UNION ALL` returns consistent counts across all dims.

### Review notes (4 Codex rounds to LGTM)
Addressed: honesty-rule fallthrough, boot-race reconciliation, runtime coverage handshake, node_bits/masks readiness decoupling, single-query concurrency, and `R2_BASE` SQL-injection hardening.

### Not in this PR
- **Publishing `sample_facet_index` to R2** (built/validated locally) — must land with/just before this so deployed multi-filter counts compute (else they honestly show "—").
- **Cold/warm DuckDB-WASM benchmark** — the release gate (Phase 4).
- Viewport + search counts (Phase 3); Playwright + remove baseline fallback (Phase 4).

Refs #305 #304 #306

### DIFF
--- a/PLAN_305_facet_counts.md
+++ b/PLAN_305_facet_counts.md
@@ -1,0 +1,128 @@
+# Plan — #305 Facet-count correctness & coherence (counts beyond the single-filter cube)
+
+**Status:** plan of record · **Authors:** Claude + Codex (joint, 2026-06-19) · **Tracks:** #305 (meta), #304 (live bug), #306 (data bug), #276 (semantics), #230, #286
+
+## Problem
+
+At the global/zoomed-out view, facet **counts** only cross-filter for a **single** active
+filter (served by the precomputed `facet_tree_cross_filter` cube via `effectiveSingleFilter()`,
+`explorer.qmd:3374`). With **2+ active selections** the cube is skipped and tree-dim counts
+revert to the **unfiltered baseline** (`explorer.qmd:3557-3562`) — deliberately, to avoid a
+~39M-row `sample_facet_membership` GROUP BY that stalls DuckDB-WASM. So the **map is correctly
+filtered** (#299 bitmask) but the **count numbers are wrong** for multi-filter at world view.
+Zoomed-in is already correct (live membership path, `explorer.qmd:3564-3598`). Reported by
+Eric in #304 (e.g. *Material: Anthropogenic* + *Specimen Type: Artifact*).
+
+## Core idea — count per node FROM the bitmasks (no membership scan)
+
+For a count dimension D over a qualifying pid set Q:
+
+```sql
+SELECT nb.concept_uri AS value, COUNT(*) AS count
+FROM <complete_facet_index> s
+JOIN read_parquet('${node_bits_url}') nb ON nb.facet_type = '<D>'
+  AND (s.<D>_mask & (1::BIGINT << nb.bit_index)) <> 0
+WHERE <AND of the OTHER active dims' mask predicates> <source> <viewport> <search>
+GROUP BY nb.concept_uri
+```
+
+- Masks are **one row per pid** → `COUNT(*)` == `COUNT(DISTINCT pid)` (do **not** add `COUNT(DISTINCT)` — wasted work).
+- Membership encodes **ancestor closure** → child bit ⇒ parent bit, so **parent ≥ child** holds for free.
+- The bitwise join is against `facet_node_bits` (≤63 rows/dim), not the 39M-row membership table.
+
+**Cross-filter semantics (confirmed correct):** OR among selected nodes *within* a dimension;
+AND *across* dimensions; **omit the target dimension's own predicate** (exclude-self), matching
+the UI contract (`explorer.qmd:3305`) and the existing cube (`build_frontend_derived.py:460`).
+Special case: **zero selected sources** ⇒ impossible filter for tree targets, but the source
+dimension's own histogram must still exclude source and stay constrained by the three trees.
+
+## Feasibility (Codex native DuckDB benchmarks, real 6M-pid data)
+
+| Query | Time |
+|---|---|
+| One unfiltered 22-node histogram | ~0.17s |
+| All three tree dims via one 56-node join | ~0.69s |
+| Eric's Material+Specimen-Type → context histogram | ~0.02s |
+| Embedded source predicate/group | ~0.014s |
+| `facets_v3` semi-join (rejected approach) | ~0.145s |
+
+Strongly supports feasibility. **Worst case ≈ 6M × 56 = ~336M cheap bit-tests**, not merely a
+6M-row scan → **WASM cold+warm benchmarking remains a hard release gate** (the app has documented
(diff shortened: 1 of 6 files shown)


rdhyee:
## ✅ Deployed to staging — how to verify

This is live on the **rdhyee fork staging** build (not org prod yet):
`https://rdhyee.github.io/isamplesorg.github.io/explorer.html`

> ⚠️ Static deploy — if you had the page open, **hard-refresh** (Cmd/Ctrl-Shift-R). First load is slow (~10–15s) while the data files download.

### The fix — two filters at world zoom (the #304 case)
1. Open this deep-link (Material = *Anthropogenic* **AND** Object type = *Artifact*, zoomed out):
   [https://rdhyee.github.io/isamplesorg.github.io/explorer.html?material=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterial%2F1.0%2Fanyanthropogenicmaterial&object_type=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterialsampleobjecttype%2F1.0%2Fartifact#v=1&lat=10.0875&lng=0.0000&alt=18604007](https://rdhyee.github.io/isamplesorg.github.io/explorer.html?material=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterial%2F1.0%2Fanyanthropogenicmaterial&object_type=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterialsampleobjecttype%2F1.0%2Fartifact#v=1&lat=10.0875&lng=0.0000&alt=18604007)
2. Look at the **Object type → artifact** count. It should read **145,770**. ✅
   - Before this fix it showed **261,086** — the unfiltered artifact total, ignoring the Material selection (that's the bug).

### Sanity anchor — one filter only
Open Material = *Anthropogenic* alone:
[https://rdhyee.github.io/isamplesorg.github.io/explorer.html?material=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterial%2F1.0%2Fanyanthropogenicmaterial#v=1&lat=10.0875&lng=0.0000&alt=18604007](https://rdhyee.github.io/isamplesorg.github.io/explorer.html?material=https%3A%2F%2Fw3id.org%2Fisample%2Fvocabulary%2Fmaterial%2F1.0%2Fanyanthropogenicmaterial#v=1&lat=10.0875&lng=0.0000&alt=18604007)
- Headline reads **157,487** (all located anthropogenic samples). This is *expected* and different from 145,770 — one filter vs. two.

### The honesty rule (optional)
With both filters active, the counts should be real numbers, **never** a silent wrong baseline. If the index ever can't answer, you'll see an explicit dash **"(—)"** instead of a misleading number — by design.

### Ground truth (computed directly from the published parquet)
| Query | Count |
|---|---|
| Total located samples | 6,026,242 |
| Material = anthropogenic only | 157,487 |
| Object = artifact only (the old buggy baseline) | 261,086 |
| **Material = anthropogenic AND Object = artifact** | **145,770** ← the fix |

*— 🤖 verification steps drafted by CC; numbers computed against production data; RY reviewing*

Issue #312, opened 2026-06-26
Make it possible to download CSV of Samples Table


akthom:
We should be able to download a csv containing all the data from the Samples Table. There should be a button at the top of the table that says "download" and when you click it you should be able to download the tbale.

rdhyee:
> 🤖 **rbotyee** (Claude, operated by @rdhyee) — *closing: shipped.*

CSV download of the samples table shipped 7/10 (Download CSV button, includes the PID + Source URL columns) and is live on production. Reopen if the export is missing anything you need. — rbotyee 🤖 (Claude, for @rdhyee)

Issue #226, opened 2026-05-21
Explorer: unify click semantics — click opens detail card, external link is one-hop-away


rdhyee:
## Problem

Two surfaces let you select a sample (map dot, table row), but their click semantics diverge — and the table-row case is internally inconsistent depending on *which pixel* you hit.

### Current behavior (`explorer.qmd` @ commit `a4f1155`)

**Map dot click** (line ~1215): runs `updateSampleCard(meta)` → side panel updates → map flies to point. **Never opens external link.** ✅ Already the right model.

**Table row click** (line 1586-1588): inconsistent based on click target.

```js
tr.addEventListener('click', async (e) => {
    if (e.target.tagName === 'A') return; // ← bleed-through
    // ... else updates side panel + flies map
});
```

The label is rendered as `<a href={sourceUrl(pid)}>{label}</a>` (line ~1558). So:

| Click target inside the row | Result |
|---|---|
| The title text (it's an `<a>`) | Handler bails → browser follows link → **external site opens** |
| Lat cell, lon cell, place column, blank space | Handler runs → side panel + map fly |

That's not random — it's "click the underlined-blue title = leave the app, click anywhere else = update the app." The most-attractive click target does the least-useful thing.

**Hover on map dot** (line ~1196): a Cesium label tooltip showing just `label || pid`. No metadata. No thumbnail. No path to the rest of the record.

## Proposed semantics

> **Click = open detail card.** The external link to the source record (OpenContext, SESAR, etc.) is a one-hop-away action **inside the card**, never the default. This applies uniformly to map-dot click and table-row click.

This matches Hana's Figma mockup ([node 222:456](https://www.figma.com/design/Nqkuqh3Z4aqVh0nmwUAgKg/iSamples-Wireframe-1.0?node-id=222-456), card subnode `225:1700`): the card title *is* the link to the external record — and the card is the click target's destination.

## Scope for the companion PR

Two changes ship together:

### Change 1 — Fix table-row link bleed-through (5 lines)
Either:
- Remove the `<a href>` wrapper from the table title cell (link moves into the card), **or**
- Replace `if (e.target.tagName === 'A') return;` with `e.preventDefault(); e.stopPropagation();` so the title click bubbles to the row handler instead of navigating.

### Change 2 — Build the in-map anchored detail card (Figma 222:456 / 225:1700)
- Card anchored near the clicked dot, with viewport-edge collision avoidance.
[...]

PR #328, opened 2026-07-10
#323: PID + Source URL columns in samples table and CSV export

### LINKED ISSUE #323: Add columns to table and csv download.
I suggest we add two more columns to the samples table under the interactive explorer. These should also be including in any csv downloads.
1) there should be a column called "Source URL" that links to the sample's landing page in its source repository
2) There should also be a column called "PID" that lists the sample's identifier. This already appears in the CSV download, but it should be in the samples table as well.

rdhyee:
## What this does

Implements @akthom's #323:

1. **PID column** in the samples table (was CSV-only) — placed after Label, ellipsized via CSS (`ark:/28722/…` pids are long) with the full pid on hover, and always complete in the CSV.
2. **Source URL column** — a dedicated external link (`n2t.net` resolver via the existing unit-tested `sourceUrl()` helper) to each sample's landing page at its source repository, in both the table and the CSV. In the CSV it's appended as the **last** column so #312-era consumers keep their column positions.

Design notes:
- Does **not** reopen #226 (which removed the label cell's href): the external link lives in its own honestly-labeled cell, and `activateRow` now ignores clicks originating inside an `<a>`, so a link click opens the record without also selecting/flying the row.
- Row-specific `aria-label` on each link (Codex review) so screen readers announce which record each identical "record ↗" link opens.

## Codex review (5.6)

> approve after one focused browser smoke test … It reuses the existing tested `sourceUrl()`; attribute values are escaped; CSV ordering remains backward-compatible; anchor clicks no longer trigger row selection; `rel="noopener noreferrer"` is correct. I found no code blocker.

## Smoke test performed (fork preview, fresh isolated browser context)

- Headers render `…|PID|…|Source URL`; pid cell ellipsizes with full pid in `title`; link `href=https://n2t.net/<pid>`, `target=_blank`, `rel=noopener noreferrer` ✅
- **Link click does NOT select/fly the row**; normal cell click still does (regression check) ✅
- Link is keyboard-focusable ✅
- Narrow viewport (480px): table scrolls inside `.table-scroll`, body does not overflow (fresh load; matches prod behavior) ✅
- 0 console errors; 48/48 unit tests ✅

Closes #323.

https://claude.ai/code/session_01AjnkWb4HpuLeDbYfzmY3X9

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -540,6 +540,26 @@ format:
         white-space: nowrap;
     }
     .samples-table tr:last-child td { border-bottom: 0; }
+    /* #323: PID column — pids are long (ark:/28722/…); keep the column
+       compact with ellipsis, full pid available on hover (title attr) and
+       always complete in the CSV export. */
+    .samples-table td.pid-cell {
+        max-width: 150px;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
+        font-size: 11px;
+        color: #555;
+    }
+    /* #323: Source URL column — dedicated external link (see activateRow's
+       anchor guard: clicking it must not also select/fly the row). */
+    .samples-table a.row-source-link {
+        color: #1565c0;
+        white-space: nowrap;
+        text-decoration: none;
+    }
+    .samples-table a.row-source-link:hover { text-decoration: underline; }
     .table-badge {
         color: white;
         padding: 2px 7px;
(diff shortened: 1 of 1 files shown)


rdhyee:
**Codex 5.6 verdict** (independent review): *"approve after one focused browser smoke test… reuses the existing tested sourceUrl(); attribute values are escaped; CSV ordering remains backward-compatible; anchor clicks no longer trigger row selection; rel=noopener noreferrer is correct. I found no code blocker."* The requested smoke test was performed (see PR body: link-vs-row click, keyboard focus, narrow viewport, fresh context) and its aria-label suggestion applied in 9ed195c.

— rbotyee 🤖

Issue #345, opened 2026-08-06
Explorer downloads ~74 MB cold: DuckDB-WASM 1.24.0's HEAD+Range probe expects 206; our (correct) 200 defeats range requests


rdhyee:
🤖 **rbotyee** (Claude, operated by @rdhyee) — *Raymond's intent:* overnight work on making isamples.org more robust at scale. *What I did:* measured the Explorer across connection speeds and browsers, root-caused a large cold-load cost, and proved the cause with a controlled A/B. Filing because the fix lives in the Cloudflare Worker, not this repo.

## Summary

A cold load of the Interactive Explorer transfers **~74 MB** before the facet panel appears. Our own docs claim *"typically less than 1 MB for initial exploration"* (index.qmd) and *"only the bytes you need are transferred"* (explorer.qmd). Both are wrong by roughly two orders of magnitude.

**The whole thing comes down to one status code.**

DuckDB-WASM decides whether a server supports partial downloads by sending a `HEAD` request with a `Range` header. `data.isamples.org` answers **`200`** instead of **`206`**, so DuckDB concludes the server can't do partial reads and downloads **every file whole** — including `samples_map_lite_v3.parquet` (62.9 MB) when it actually needs about 1.5 MB of it.

## Proposed fix

In the Cloudflare Worker fronting `data.isamples.org`: when a `HEAD` carries a `Range` header and the object supports ranges, respond **`206`** with `Content-Range` (no body, per `HEAD` semantics) rather than `200`.

`Content-Range` is already in `Access-Control-Expose-Headers`, so nothing else needs to change.

## Evidence — controlled A/B, not inference

A transparent reverse proxy forwarded every request to the real host unchanged, altering **exactly one thing**: `HEAD`+`Range` → `206`. Same build, same cold cache, one variable.

**Unthrottled**

| | Control (`200`) | Treatment (`206`) |
|---|---:|---:|
| Bytes from data host | **74,202,598** | **3,341,812** |
| `full HTTP read` fallbacks | 8 | **0** |
| `samples_map_lite_v3.parquet` | 62,924,115 B (whole file) | 1,467,731 B (8 ranged reads) |
| `sample_facet_masks.parquet` | 10,138,648 B (whole file) | 767,000 B (48 ranged reads) |

**3G fast (1.6 Mbps) — the user-visible result**

| | Control | Treatment |
|---|---:|---:|
| Globe drawn | 54.0 s | 54.1 s |
[...]

Issue #340, opened 2026-08-04
Text search (still) doesn't update facet counts correctly


ekansa:
I tried a text search for the suggested "pottery Cyprus": https://isamples.org/explorer.html?search=pottery+Cyprus#v=1&lat=35.3300&lng=33.0659&alt=200000&mode=point

It says that it found a total of 1305 results, but the counts for various facet options typically greatly exceed the total number of search results. The facet counts need to be recalculated based on the full-text search filter.

rdhyee:
🤖 **rbotyee** (Claude, operated by @rdhyee) — *Raymond's intent:* get a cheap, honest fix in before the grant closes, and tell Eric straight what we found. *What I did:* reproduced #340, tracked down the cause, shipped a fix in #341 (now live on isamples.org), and wrote it up here.

Thanks Eric — confirmed, and it's a bit worse than it looked. Those facet numbers weren't just too high, they were the **whole-database totals**, completely ignoring your search. Every one of the 1,305 matches is OpenContext, but SESAR was still sitting there claiming 4,389,231.

## What's actually going wrong

It's not that we forgot to apply the search to the counts. That part exists. The problem is that the recalculation **starts and then never finishes** — and while it's running, the numbers on screen just go slightly greyed out. The text itself never changes. So the old, wrong numbers stay up looking perfectly normal, indefinitely.

That's why this was easy to miss: there's no error, no spinner, no blank. It just quietly keeps showing you the last answer it had. I watched it sit there unfinished 15+ seconds after the search was done, and moving the map around didn't jog it loose either.

## What we've done about it (#341, deployed)

Making the counts actually respect a search is the unfinished half of the #305 plan, and it isn't something we can responsibly finish and test in the time left.

So instead of leaving wrong numbers on screen, **a search now shows `(—)` for every count** — an honest "we don't know" rather than a confident wrong answer. That's your own request from #304, taken literally: *"make sure the user interface does NOT show inaccurate numbers."*

Clear the search and the real numbers come straight back. Filtering by facets on its own is untouched and still works normally.

Verified on the live site just now: with `?search=pottery+Cyprus` all 60 count elements show the dash and none are stuck mid-recalculation; with a facet filter and no search, real cross-filtered counts still appear (SESAR `0`, OpenContext `261,086` for Artifact).

## I'm deliberately leaving this issue open

[...]

PR #347, opened 2026-08-06
DRAFT (do not merge): coherence F2 — invalidate facet counts at input, before the await

rdhyee:
> ## ⚠️ DRAFT — do not merge. Codex has blocked this twice and I agree with it.
>
> Opened as a draft so the work is visible and reviewable, **not** because it's ready.
> Built overnight; two substantive blockers remain, detailed at the bottom.

## What this is

The coherence audit's highest-value open item (F2), and the approach Codex argued for over a watchdog: *"the invariant belongs at input invalidation, not at a timer attached to selected query paths."*

## The bug it targets

Both filter handlers had this shape:

```js
writeQueryState(); refreshHeatmap();
await <globe reload>;          // long
refreshFacetCounts();          // only NOW are counts invalidated
```

That await is not short. Measured overnight on production: **168 s on 4G, 423 s on 3G** — because the Explorer currently downloads ~74 MB on cold load (separate finding, [#345](https://github.com/isamplesorg/isamplesorg.github.io/issues/345)). For that entire window the facet counts kept displaying the **previous filter's numbers, unmarked**.

#340 was this same shape with a different trigger. #341 and #342 fixed instances; this targets the class.

## The change

`invalidateFacetCountsNow()` — synchronous, called before any await: bumps the request id, clears the pending debounce, and marks counts recomputing (which swaps text to `(Loading…)` after 400 ms via #342). It deliberately does *not* schedule a recompute. An inner `try/finally` guarantees a successor on success **and on throw**.

## Demonstrated effect

400 kbps, toggle one source filter, sample the DOM every 0.4 s:

| | production (no fix) | this branch |
|---|---|---|
| invalidated within 0.4 s | False | True |
| stale **unmarked** windows | **27** | **0** |
| what the user sees | `(4,389,231)` for 10+ s after deselecting SESAR | `(Loading…)` |

## Why it is NOT ready

**1. Overlapping handlers can still fire a premature query.** Change A invalidates and awaits; change B invalidates and starts newer work; A is superseded, settles first, and its `finally` schedules a recompute that can run while B's globe reload is still pending. `facetCountsReqId` prevents a stale *repaint* but not the premature *query*. Needs a dedicated filter-invalidation generation — not `facetCountsReqId`, which camera refreshes also mutate.

**2. The verifier cannot prove what I claimed.** It only ever lets the awaited work *succeed*, and the pre-fix code also repainted on success — so the run cannot distinguish the new `finally` from the old placement. It needs deterministic fault injection (force the query-state write to throw right after invalidation) and per-element rather than aggregate assertions.

**3. Also outstanding:** `applySearchFilterChange()` should get guaranteed-successor placement (not pre-await invalidation) — after clearing a search, a rejected reconcile can strand the old dash even though counts are computable again.

**4. Untested risk:** `facet-tree.spec.js` uses fixed 3 s sleeps then parses numeric text; if a globe reload exceeds 3 s it could now read `(Loading…)`. I could not test this — the spec needs a `docs/data` mirror absent from this checkout. I confirmed it fails **identically** on a control build without this change (10/10 both), so this branch doesn't regress it, but the specific concern is unverified.

## Verification that did pass

`quarto render` clean · `tests/test_smoke.py` PASS · `tests/test_frontend_derived.py` 40 passed · #341/#342 unaffected (search → 60/60 `(—)`, 0 stuck; facet-only → real counts) · no page errors · deployed to fork staging.

Refs #340, #304, #305

### DIFF
--- a/explorer.qmd
+++ b/explorer.qmd
@@ -4555,10 +4555,15 @@ zoomWatcher = {
     function invalidateFacetCountsNow() {
         clearTimeout(facetCountsDebounce);
         ++facetCountsReqId;
-        // Under an active search the counts are already the honest "(—)" dash
-        // (#340). Re-dimming them would downgrade "we cannot know this" to
-        // "we're about to know it", which is a worse claim, not a better one.
-        if (searchIsActive()) return;
+        // Under an active search the counts are genuinely unavailable (#340), not
+        // pending. Re-dimming them would downgrade "we cannot know this" to
+        // "we're about to know it" — a worse claim, not a better one. Repaint the
+        // dash rather than merely returning, so this is idempotent and does not
+        // depend on an earlier caller having already painted it (Codex review).
+        if (searchIsActive()) {
+            markFacetCountsUnavailable();
+            return;
+        }
         markFacetCountsRecomputing();
     }
 
@@ -5084,6 +5089,14 @@ zoomWatcher = {
             // can run for minutes on a slow link. Without this the facet counts
             // keep asserting the pre-toggle numbers for that whole window.
             invalidateFacetCountsNow();
+            // The inner finally GUARANTEES a successor (Codex review). Invalidating
+            // bumps facetCountsReqId and cancels the pending debounce, so if
+            // anything between here and the end of the globe work throws, nothing
+            // would ever schedule a recompute and the counts would sit at
+            // "(Loading…)" until some unrelated later event. A hung promise
+            // staying "Loading" is honest (#342); a REJECTED one is terminal, so
+            // "Loading" would then be a lie of a different kind.
+            try {
             updateSourceLegendState();
             writeQueryState();
             refreshHeatmap();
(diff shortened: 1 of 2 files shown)


rdhyee:
🤖 **rbotyee** (Claude, operated by @rdhyee) — *Raymond's intent:* close out the project's stale open PRs. *What I did:* closing this one, on Raymond's approval, for the reasons below.

Closing rather than merging or rebasing. In your own words above: "Codex has blocked this twice and I agree with it" — and the two correctness gaps you documented (a race between overlapping filter handlers that can still fire a premature query; a verifier that can't actually distinguish this fix from the pre-fix code) are still unresolved. The problem this targets — facet counts going stale under a filter change — has a shipped, documented stopgap live in production (#341/#342: an honest `(—)` instead of a wrong number), so nothing regresses by closing this.

The diagnostics here (the invalidation-generation gap, the fault-injection testing gap) are good and worth keeping as a reference — this comment links back to them for whoever picks up #340's real fix next, rather than carrying a known-not-ready branch open indefinitely.

— isamples-0914 (Claude, for rdhyee)

Issue #313, opened 2026-06-26
Need to explore what browser/operating system/bandwidth combinations the Explore works and does not work in


akthom:
We are observing some performance differences for the Explorer that might be due to differences in browser/operating system/bandwidth combinations. For instance, I'm using Firefox on a macbook, and the Explorer doesn't seem able to load some searches that are able to load on other people's systems. here's a screen shot: 

<img width="1443" height="456" alt="Image" src="https://github.com/user-attachments/assets/72ff706c-4ef2-465e-9b7a-417cc197e9e8" />

When I try to combine two material facets, the explorer chokes and just displays "--" next to the facet counts. But Eric and Raymond were both able to load this search and get correct counts.

Could you explore whether there are browser/operating system combinations that can't run the explorer efficiently? And if there's a recomended minimum amount of bandwidth/ internet connectivity needed to run the explorer? And recommend what browser the explorer works best in? users should be able to click multiple facets on and off without the explorer crashing. They should be able to spin the globe around without it breaking.

rdhyee:
## Investigation: the "--" is the honesty-dash; root cause is a boot-time loading race

Thanks @akthom — we traced this. The "**--**" you see isn't a crash; it's a deliberate **"honesty" signal** added in the recent facet-count work (#304/#305/#308): when the Explorer can't yet compute correct cross-filtered counts, it shows a dash instead of a *wrong* number. So your dash means the count-index wasn't ready (or failed to load) at the moment you applied the 2nd material facet.

### Most likely root cause — a boot race (slow connection / Firefox)
Before multi-filter counts work, the page must download **and scan ~20 MB** of count data at boot (`sample_facet_masks` 9.7 MB + the new `sample_facet_index` 9.7 MB, the latter via a **full-table scan** coverage check). On a modest or throttled connection — or a backgrounded Firefox tab, where network/timers are throttled — that takes **20–80+ seconds**. If you select the 2nd facet inside that window, the count path returns "not ready" → the dash.

- It **auto-recovers** once loading finishes (counts fill in on their own) — *unless* the preflight throws (network error / DuckDB-WASM hiccup), in which case the dash is permanent for that session and a refresh is needed.
- This compounds with **#190** (DuckDB-WASM 1.24.0 sometimes "falls back to a full HTTP read" instead of fetching byte-ranges), which is worse in Firefox Workers.

[...]

PR #348, opened 2026-08-06
#345: narrow HEAD+Range shim so DuckDB-WASM uses range requests (74 MB -> 3 MB cold load)

rdhyee:
Fixes the 74 MB cold load in #345 — but read the framing, because this is a **deliberate standards divergence**, not a bug fix.

## What changes

DuckDB-WASM 1.24.0 probes for range support with exactly `HEAD` + `Range: bytes=0-` and requires a `206`. We answer `200`, so it logs *"falling back to full HTTP read"* and downloads whole files — 62.9 MB of `samples_map_lite_v3.parquet` when it needs ~1.5 MB.

This makes **only that exact probe** return `206`. Every other ranged `HEAD` stays `200`.

## Why it's framed as a shim

RFC 9110 §14.2: `Range` is defined only for `GET`; servers **MUST ignore** it on other methods. **Our `200` was correct.** DuckDB's probe is the nonconforming party. So this is a compatibility measure with a named removal path — delete it when the Explorer stops using Quarto's pinned duckdb-wasm and does its own init on a conformant version.

Codex talked me out of calling this a fix, and out of the wider version I first wrote (which would have answered 206 to *any* ranged HEAD).

## Evidence

Controlled A/B — a transparent proxy changing only this one behaviour:

| | Control | Treatment |
|---|---:|---:|
| Cold bytes | 74,202,598 | 3,341,812 |
| Full-read fallbacks | 8 | 0 |
| Facet panel on 3G | 440.6 s | 94.2 s |
| **Warm bytes (relaunch)** | **0** | **1,467,731** |
| Warm time-to-facets | 4.2 s | 4.2 s |

So: **~71 MB saved on first visit, ~1.47 MB cost per repeat visit**, no warm time regression. The warm number is included because it could have reversed the recommendation — and the original issue omitted it.

## Test coverage

`test/range_contract.sh` — 23 assertions against `wrangler dev --local` with a seeded local R2:

- Unfixed Worker: **3 failures**, all `HEAD`+`Range`
- With this change: **23/23**

Four assertions exist specifically to prove the shim **doesn't widen** (`bytes=0-99`, `bytes=100-199`, `bytes=-100` must all stay `200`). Caching, CORS, ETag-stability and ranged `GET` are all asserted unchanged — this Worker exists for caching, so that contract is guarded.

## ⚠️ How to test this WITHOUT touching production

**There is no staging data host.** This Worker serves `data.isamples.org` on a route, so a plain `wrangler deploy` goes straight to production for every consumer.

Safe path — deploy to a workers.dev URL instead, then point the staging Explorer at it:

```bash
cd workers/data-isamples-org

# 1. Deploy to a NON-production name, with NO route.
#    Comment out the `routes = [...]` block in wrangler.toml first,
#    or use a separate wrangler config with a different `name`.
npx wrangler deploy --name isamples-data-345-canary

# 2. Point the fork's staging Explorer at the canary (no rebuild needed —
#    the Explorer supports a data_base override):
#    https://rdhyee.github.io/isamplesorg.github.io/explorer.html?data_base=https://isamples-data-345-canary.<subdomain>.workers.dev

# 3. Verify the handshake actually changed:
curl -sI -H 'Range: bytes=0-' https://isamples-data-345-canary.<subdomain>.workers.dev/isamples_202608_h3_summary_res4.parquet
#    expect: HTTP/2 206 + content-range

# 4. Measure end-to-end (harness from #346):
python tests/playwright/bandwidth_matrix.py https://rdhyee.github.io/isamplesorg.github.io \
    --profiles unthrottled,3g-fast --budget 600 \
    --query "?data_base=https://isamples-data-345-canary.<subdomain>.workers.dev"
```

Only promote to the `data.isamples.org` route once the canary shows 206s and reduced bytes on the real edge.

## What is NOT verified

Everything measured here is a **local** `wrangler dev` Worker plus a Python proxy over `http://localhost` — **not the real Worker on the real Cloudflare edge**. That canary is the gap.

Also relevant: Workers Caching appears **disabled** today (production returns no `CF-Cache-Status`, and the Worker demonstrably still sees `Range`). If it is ever enabled, Cloudflare strips `Range`, caches a full `200` and slices its own `206`s — which would replace this path entirely and require re-testing.

Refs #345, #313

### DIFF
--- a/tests/test_data_origin_contract.py
+++ b/tests/test_data_origin_contract.py
@@ -1,0 +1,172 @@
+"""
+HTTP contract tests for the data origin (`data.isamples.org`).
+
+WHY THIS FILE EXISTS
+--------------------
+The Explorer's whole architecture rests on one assumption: DuckDB-WASM fetches
+only the byte ranges a query touches. In August 2026 we discovered that had been
+silently false — a cold load transferred **~74 MB instead of ~3 MB** because
+DuckDB's range-support probe was answered in a way it did not accept, so it fell
+back to downloading whole files (#345).
+
+That regression was invisible to every existing test. It was also *mis-cleared*
+in June (`ISSUE_313_FINDINGS_2026-06-26.md`) by a `curl` probe that used **GET**
+where DuckDB uses **HEAD** — the check looked right and proved nothing.
+
+These tests are deliberately cheap: no browser, no DuckDB, a handful of bytes on
+the wire. They run before the Playwright smoke gate so this class of failure is
+caught before anything downloads 74 MB to discover it.
+
+MAINTENANCE NOTE
+----------------
+`tools/build_release_manifest.py` performs the ranged-GET probe but NOT the HEAD
+probe — which is exactly why it could report a healthy origin while DuckDB was
+broken. If you touch either, keep them in sync.
+"""
+import json
+import os
+import pathlib
+
+import pytest
+import requests
+
+ORIGIN = os.environ.get("ISAMPLES_DATA_ORIGIN", "https://data.isamples.org")
+PAGE_ORIGIN = "https://isamples.org"
+MANIFEST = pathlib.Path(__file__).resolve().parent.parent / "isamples_202608_release_manifest.json"
+
+# The Worker 403s some default user agents; identify honestly.
+UA = {"User-Agent": "isamples-ci-contract/1.0 (+https://isamples.org)"}
+TIMEOUT = 30
+
+
+def _manifest_files():
+    if not MANIFEST.exists():
+        pytest.skip(f"release manifest not found at {MANIFEST}")
+    return json.loads(MANIFEST.read_text())["files"]
+
+
+def _boot_critical_large_file():
+    """The biggest boot-critical parquet — the one a full read actually hurts.
+
+    Skips (rather than fails) if the origin does not serve it. Without this a
+    404 from, say, a partially-seeded local test origin gets reported as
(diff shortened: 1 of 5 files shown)


rdhyee:
🤖 **rbotyee** (Claude, operated by @rdhyee) — *Raymond's intent:* get #345 over the finish line, carefully. *What I did:* three Codex rounds over the deploy path (read-only; no Cloudflare writes), fixed every finding, applied `CC+Codex+LGTM`. **Raymond runs the deploy himself; nothing has been deployed to production by this PR yet.**

**Codex verdict at `df89db3` — LGTM.** Rounds: (1) failed-deploy ≠ unchanged → re-query + verify + always print rollback; version-id parsing fail-closed; probe timeouts; contract test wired into CI. (2) `--runxfail` so the CI gate can actually fail; a duplicated timeout. (3) clean.

**Evidence, all 2026-08-27:** canary passes every check incl. Content-Length; staging Explorer against the canary: 0 "falling back" warnings, identical counts; production pre-flight via `deploy-prod.sh --preflight` passes read-only; active production version `b61d3bbc-…` (Worker source unchanged on `main` since → production receives exactly the shim); bytes measured three ways (see the corrected claim in the description and #351); background reads shown to terminate (120 s = 60 s).

**Sequence from here:** `./deploy-prod.sh --preflight` → `./deploy-prod.sh` (RY) → live Explorer console check + contract tests (CC) → remove the `xfail` marker (commit to this branch) → merge → `./deploy-canary.sh --teardown`.

